PR #23, opened 2026-08-11
docs: cut the agent context surface to a verified core

KitsuneKode:
## What changed

Reduced the agent-facing doc surface from 374 tracked files to a verified core, fixed the stale facts that were misrouting agents, and added two checks so the rot that produced them fails the build instead of accumulating silently.

## Checklist

- [ ] Changeset added (`bun run changeset`) — **N/A, docs-only**
- [ ] `bun run guard` — **N/A**, `apps/cli/package.json`, changelogs, and `.changeset/**` are untouched
- [x] `bun run fmt && bun run lint && bun run test && bun run typecheck` — fmt, lint, typecheck, and `fmt:check` green. See the test note below.
- [ ] `bun run build` — **N/A**, no feature, playback, provider, release, or packaging code changed
- [x] Live provider and Discord smokes skipped intentionally — nothing in this diff touches provider or presence runtime

### Test note

The suite is unchanged at **3698 pass / 1 fail**, identical before and after this branch. The one failure is `test/unit/services/update/native-installer/rollback.test.ts` — it `chmod`s a directory to `0o555` and expects the write to fail, which cannot hold when the suite runs as root. It reproduces on a clean checkout of `main` and passes for any non-root user. Not caused by, and not fixed by, this PR.

---

## Why

The entry point was routing agents to files that did not exist — `CONTEXT.md`, `.docs/experience-overview.md`, `.docs/product-prd.md`, `plan-implementation-truth.md` — and stated three invariants that were false:

- the provider registry had moved to `container/bootstrap-providers.ts`
- `apps/cli/index.ts` had been deleted
- config paths are platform-resolved by `getKunaiPaths()`, not always `~/.config`

The pattern behind most of it: `apps/cli/src/app/` was reorganized into subdirectories and no doc followed. `PlaybackPhase.ts`, `SessionController.ts`, and all five share-link modules were cited at their old flat paths across five docs.

Meanwhile the rules that actually fail the build — the layering and package-DAG gates in `apps/cli/test/unit/architecture/` — appeared nowhere in the agent surface.

## Entry point

`AGENTS.md` (symlinked as `CLAUDE.md`) rewritten around routing rather than prose. Every link resolves. Non-negotiables now lead with what is machine-checked.

## New docs

- **`.docs/feature-map.md`** — routes any user-visible feature to the directory that owns it, including a "not in the active runtime" section so nobody cites `cineby`, `vidrock`, or `archive/legacy` as capabilities.
- **`.docs/telemetry-privacy-contract.md`** — opt-in and payload invariants, verified against `services/telemetry` and `apps/telemetry-ingest`.

## Pruning

Every tracker now has the same contract: the top level holds unfinished work, `archive/` holds history and carries no authority, and each archive README explains how to recover work from it without reviving the file.

| Surface | Live | Archived |
| --- | --- | --- |
| `.docs` | 31 | 30 |
| `.plans` | 44 | 99 |
| `plans` | 19 | 13 |
| `docs/superpowers` | 9 | 71 |

Notable: `plan-implementation-truth.md` (81 KB) was archived rather than slimmed — once `.plans` holds only unfinished work, adjudicating plan-vs-code drift has nothing left to adjudicate, and its first entry was itself stale. `roadmap.md` was a changelog; it is now an index.

A competing verification checklist had accumulated at `docs/superpowers/plans/END-OF-RUN-VERIFICATION.md` despite `VERIFICATION.md` claiming sole ownership. Its open items had already shipped — it asks for a `mediaLanguageProfileFor(state)` helper that now exists in `domain/media/content-kind` with two callers — so it was archived rather than merged.

## Documented invariants that were missing

`AGENTS.md` had claimed six enforced rules. There are eleven, across four test files. The omissions were the ones that bite:

- **`apps/cli/src` root is closed** — only 14 allowlisted flat modules; a new `.ts` file there fails the build
- **`packages/*` follows a strict DAG** with a per-package allowlist (`core` does not depend on `schemas`; `design` depends on nothing)
- `app` and `infra` must not import provider packages; `app-shell` must not import `@kunai/storage`
- `filename-convention.test.ts` and `dev-entrypoint.test.ts` were undocumented entirely

## Guardrails

- **`bun run verify:doc-paths`** — fails when a doc cites a path that is gone. Found 34 dead references on first run.
- **`bun run verify:doc-coverage`** — the inverse: every subdirectory of `services/`, `domain/`, `infra/`, `app/`, and `packages/` must be routed in the feature map. Found 30 unrouted areas, 14 of which a manual grep had missed. Mutation-tested before being trusted green.

**Known limit, stated in `AGENTS.md`:** neither check reads source comments, so a `// See docs/...` reference can still rot silently. Four source files carry such references today.

## Review guidance

The two scripts in `scripts/` are the reviewable core. Everything else is `git mv` plus repointed links — `git log --follow` works on every moved file. Source changes are limited to four comment-line repoints, two new scripts, and two `package.json` script entries.

This also closes `plans/017`, which had been sitting BLOCKED.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CksGEhZULcKbXKEH1npCcg)_


KitsuneKode:
Closing as superseded. This work already landed on `main` via merge 6d9098bf (branch `docs/agent-context-surface`) — `main` has `scripts/verify-doc-paths.ts`, the archived `plans/`, and the trimmed agent surface. This branch did the same job in parallel, which is why it now conflicts across 261 files; resolving that would mean re-landing shipped work.

One thing here never made it over: `scripts/verify-doc-coverage.ts`. That has been salvaged in #25, together with the feature-map rows it depends on — the script fails against `main` alone with 30 unrouted directories, so it could not be cherry-picked by itself.

Nothing else from this branch appears to be missing from `main`. If you spot something, reopen and we'll pull it across the same way.

PR #33, opened 2026-08-13
fix(doctor): report curl, which the default anime provider needs

KitsuneKode:
## Scope

`kunai doctor` reported `mpv`, `yt-dlp` and `ffprobe` but not `curl`.

AniDB is the **default** anime provider (`animeProvider: "anidb"` in
`packages/config/src/defaults.ts`), `anidb.app` sits behind Cloudflare, and Bun's
own `fetch` is frequently challenged there — a challenge page parses to zero
search results. So a host without curl could get an empty anime search with no
diagnostic on any surface.

## What changed

- `CapabilitySnapshot` gains `curl: boolean`.
- A `curl-missing` capability issue, `degraded` severity — anime is one mode, so
  it degrades that route rather than blocking the shell. The message names AniDB,
  because "curl not found" alone gives the user no reason to care.
- Reported in the doctor summary line, a setup dependency row, and the capability
  fingerprint.

### The probe asks the provider instead of guessing

`resolveAnidbCurl()` already owns the candidate list and prefers a
curl-impersonate build (`curl_firefox135`, `curl_chrome136`, …) over plain `curl`,
since Cloudflare fingerprints the TLS handshake. Probing for the literal `"curl"`
would report *missing* on a host running only `curl_chrome136` — a false negative
contradicting what the provider can actually do. So the probe calls
`resolveAnidbCurl`, keeping one source of truth.

### Three readers, not a bare declaration

Silent no-ops are this repo's stated house failure mode, so the new flag was
wired into every surface that already reads the snapshot:

| Surface | Before | After |
| --- | --- | --- |
| `doctor` summary | `mpv=ok yt-dlp=ok ffprobe=ok` | `… ffprobe=ok curl=ok` |
| setup dependency rows | mpv / yt-dlp / ffprobe / posters | + `curl` |
| capability fingerprint | no curl bit | includes curl |

The fingerprint matters on its own: it gates whether the remediation notice
reappears, so without it a user could install curl and still be told it was
missing.

### One incidental cleanup

`setup-workflows.ts` had an inline fallback snapshot duplicating
`probeCapabilities()`, and it had already drifted. It now calls
`probeCapabilities()`. This was not optional polish — `app-shell` may not import
`@kunai/providers` (enforced by `boundary-imports.test.ts`), so that literal
could not have probed curl itself.

## Verification

All from the worktree root:

| Gate | Result |
| --- | --- |
| `bun run typecheck` | 14/14 |
| `bun run lint` | 12/12 |
| `bun run fmt:check` | clean |
| `bun run test` — unit | **3986 pass / 0 fail** |
| `bun run test` — integration | **92 pass / 0 fail / 47 skip** |
| `bun run test` — providers | 264 pass |
| `bun run test` — docs | 71 pass |
| `bun run build` | 10/10 |
| `git diff --check` | clean |
| `verify:doc-paths` / `verify:doc-coverage` | pass |
| `check-codegen-freshness.ts` | fresh |

The integration suite's 47 skips are the repository's existing baseline, not
exclusions introduced here.

New tests: `apps/cli/test/unit/capability-curl.test.ts` (6), written failing
first. They inject a PATH stub, so they assert behaviour rather than whatever the
developer happens to have installed — including the impersonate-build case that a
naive probe would get wrong.

### Observed on the built binary

```
$ kunai doctor
  mpv=ok yt-dlp=ok ffprobe=ok curl=ok

$ env PATH=/nonexistent kunai doctor
  mpv=missing yt-dlp=missing ffprobe=missing curl=missing
  [warning] curl-missing: curl not found — AniDB (the default anime provider)
  sits behind Cloudflare and needs it; anime search may return nothing without it.
      - Arch:   sudo pacman -S curl
      ...
```

## Docs

`.docs/quickstart.md` (runtime tools) and
`docs/users/diagnostics-and-reporting.mdx` (capabilities list), plus regenerated
`apps/docs/lib/generated-metadata.json`. Changeset: `patch`.

## Release blockers

None introduced or discharged. This makes an existing dependency gap
*diagnosable*; it does not change provider behaviour.

https://claude.ai/code/session_012N9Tm1cSoB5pvYjrThFvsz

### DIFF
--- a/.changeset/doctor-reports-curl.md
+++ b/.changeset/doctor-reports-curl.md
@@ -1,0 +1,5 @@
+---
+"@kitsunekode/kunai": patch
+---
+
+Report `curl` in `kunai doctor` and setup. AniDB is the default anime provider and needs a curl (plain or curl-impersonate) to get past Cloudflare, so its absence could previously make anime search return nothing with no diagnostic anywhere.
--- a/.docs/quickstart.md
+++ b/.docs/quickstart.md
@@ -11,6 +11,12 @@ Use this doc for setup, local execution, and common environment issues. Architec
 - ImageMagick (`magick`) if you want Kitty/Ghostty non-PNG poster conversion
 - `yt-dlp` for YouTube playback and downloads/offline queue (must be on `PATH` for YouTube resolve/play and when downloads are enabled)
 - `ffprobe` optional—used only for quick validation of finished files, not downloading
+- `curl` for anime mode: AniDB is the default anime provider and sits behind
+  Cloudflare, which frequently blocks Bun's own fetch. Without a `curl` on
+  `PATH`, anime search can come back empty. A curl-impersonate build
+  (`curl_chrome136`, `curl_firefox135`, …) is preferred where available, since it
+  matches a browser TLS handshake; plain `curl` is used otherwise. `kunai doctor`
+  reports this as `curl=ok` / `curl=missing`.
 
 Deeper reference for terminal graphics, env overrides, and testing: [.docs/poster-image-rendering.md](poster-image-rendering.md).
 
--- a/apps/cli/src/app-shell/setup-shell.tsx
+++ b/apps/cli/src/app-shell/setup-shell.tsx
@@ -243,6 +243,14 @@ function SystemSlide({
       install: "Install ffprobe from your platform media-tools package when needed",
     },
     {
+      name: "curl",
+      status: snapshot.curl ? "ok" : "optional-missing",
+      detail: snapshot.curl
+        ? "Anime search ready"
+        : "Needed by AniDB, the default anime provider — anime search may return nothing",
+      install: "brew install curl  ·  pacman -S curl  ·  apt install curl",
+    },
+    {
       name: snapshot.image.renderer !== "none" ? "posters" : "posters (chafa/kitty)",
       status:
         snapshot.image.renderer !== "none" ? "ok" : snapshot.chafa ? "ok" : "optional-missing",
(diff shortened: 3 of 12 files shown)


KitsuneKode:
## Windows CLI parity — external infrastructure, not this change

`Windows CLI parity` is red on this PR. It failed twice with the same root cause, and it is not related to the diff.

The `Provision native mpv` step cannot install mpv because Chocolatey's V2 feed is returning **504 Gateway Timeout**:

```
Failed to fetch results from V2 feed at
'https://community.chocolatey.org/api/v2/Packages(Id='mpvio.install',Version='0.41.0')'
with following message : Response status code does not indicate success: 504 (Gateway Timeout).
Chocolatey installed 0/0 packages.
```

mpv therefore never lands on PATH and the next line fails:

```
The term 'mpv.exe' is not recognized as a name of a cmdlet, function, script file, or executable program.
```

Evidence it is not this branch:

- This PR touches no CI workflow and nothing in mpv provisioning — it changes `kunai doctor` capability reporting.
- Both the original run and the `--failed` re-run failed at the same step with the same 504.
- `fix/offline-playback-hardening` (PR #32, unrelated branch) also has Windows failures in the same window, so the condition is repo-wide right now.

Every other check passes: Typecheck, Lint, Format check, Test, Build CLI + npm pack, linux binaries, Docs build, Docs coverage, CodeQL, GitGuardian, all installer scenarios, and macOS CLI parity.

I am explicitly **not** calling this PR all-green. This needs a re-run once the Chocolatey feed recovers, and it should not be merged on my assertion that the failure is environmental — please confirm the re-run before merging.

KitsuneKode:
Re-run after the Chocolatey feed recovered: **Windows CLI parity now passes**. All checks green, `mergeStateStatus: CLEAN`. The 504 was transient infrastructure as diagnosed above — no code change was needed.

PR #35, opened 2026-08-13
feat(poster): zero-install native poster pipeline, plus iTerm2/VS Code inline images

KitsuneKode:
Implements `docs/superpowers/plans/2026-08-11-zero-install-poster-pipeline.md`
(all 9 tasks), and adds inline-image support for iTerm2 and VS Code.

Branched from `origin/main` directly, **not** from PR 4 — that stack was dissolved
by the 2026-08-12 scope decision.

## What this does

Every poster now goes through one bounded native preparation, then to whichever
renderer the terminal actually supports:

```
bounded source  →  preparePoster (Bun.Image)  →  Kitty        ← prepared PNG
(16 MiB cap,       resize + EXIF + PNG bridge    iTerm2       ← prepared PNG
 streamed)                                       Sixel        ← prepared RGBA
                                                 Half-block   ← prepared RGBA
```

Nothing re-decodes, and **nothing needs installing** — which is what let chafa and
ImageMagick be removed rather than kept as optional upgrades.

## Every API assumption was probed before any code was written

The plan asserted a specific `Bun.Image` surface. Verified against Bun 1.3.14 first:

| Claim | Observed |
| --- | --- |
| options-object constructor | accepted |
| `fit: "inside"` | 40×20 into a 20×20 box → **20×10** (not cropped to 20×20) |
| `withoutEnlargement: true` | 2×4 source stays **2×4** |
| `maxPixels` enforced | throws `input exceeds maxPixels limit` |
| corrupt bytes | throws `unrecognised format` |
| PNG output shape | bitDepth 8, colorType 6, non-interlaced — exactly what the repo's `decodePng` accepts |
| EXIF orientation 6 | a 2×3 JPEG reports **3×2**, proving `autoOrient` works |

All held, so the plan needed no correction on this point.

## Image quality across terminals

**iTerm2 was not detected at all.** `detectTerminal` had no branch for it, so it
fell through to `unknown` and rendered half-block ASCII on a terminal with
first-class image support. VS Code was in the same position.

| Renderer | Fidelity | Terminals |
| --- | --- | --- |
| Kitty graphics | full resolution, true colour | kitty, Ghostty |
| **iTerm2 inline (new)** | full resolution, true colour | iTerm2, VS Code ≥1.80 |
| Sixel | full resolution, 64 colours | Windows Terminal ≥1.22, WezTerm, foot |
| Half-block | text approximation | everywhere, and all multiplexers |

Inline images are ordered **before** the sixel probe: iTerm2 answers the DA1 sixel
query too, and sixel is quantised while an inline image carries the prepared PNG
verbatim — taking sixel there would be a downgrade.

Three deliberate restraints:

- **VS Code is version-gated** on `TERM_PROGRAM_VERSION ≥ 1.80`. Emitting the escape
  to an older build sprays raw bytes across the UI — the same caution the repo
  already applies to Windows Terminal sixel. An unreported version is treated as too old.
- **`LC_TERMINAL` is accepted** for iTerm2, which forwards it through ssh where
  `TERM_PROGRAM` is lost. It is checked *after* the kitty-compatible names so an
  inherited value cannot outrank the terminal actually in front of the user.
- **WezTerm stays on sixel** although it also implements the protocol, because the
  sixel overlay path is the verified one there. `KUNAI_IMAGE_PROTOCOL=iterm` is
  available; promoting it to default needs a real-terminal check first.

## Correctness changes worth review

**Kitty upload and placement are now separate.** Registering the slot inside
rendering meant an upload finishing after its caller was superseded could replace a
live placement. The caller now commits only after confirming it is still current,
and the commit is synchronous and adjacent to that check. A stale upload is deleted
*directly*, never released through the registry — it owns no slot, so a registry
release would corrupt bookkeeping for whichever image legitimately holds it.

**Both acquisition paths were unbounded** (`arrayBuffer()` on remote bodies and
local sidecars). Both now stream against the 16 MiB ceiling. `Content-Length` is a
cheap early rejection only, never the defence — it is attacker-controlled, so the
stream is bounded independently and a server that understates it is still cut off.
Rejection *cancels the reader*, which is what actually stops the transfer.

**Caches are byte-budgeted, not entry-counted.** Forty half-block thumbnails and
forty full-width posters counted the same before. Source 48 MiB / 24 entries;
prepared 32 MiB / 64 entries, weighed by PNG + RGBA since both stay resident.

## Deliberate behaviour changes

1. **Undecodable-but-PNG-signed bytes no longer reach the terminal.** The old Kitty
   path recognised the signature and forwarded them undecoded. Everything is decoded
   now — that is what makes the pixel ceiling and EXIF orientation enforceable.
   Three image-pane fixtures were a bare 9-byte PNG signature relying on that
   pass-through and are now real images.
2. **`chafa-symbols` and the `symbols` protocol are gone**; those terminals get
   half-block. `ImageCapability.dependency` is gone too — it had no reader outside
   the module that set it.

## Two bugs this surfaced

- `setup-workflows` hand-rolled a `CapabilitySnapshot` still reporting `chafa` and
  `magick` after both were retired. TypeScript missed it because the `??` widened
  the type instead of checking the literal. It now calls `probeCapabilities()`.
- A capability test asserted chafa lookup was *memoized to one call*. It now asserts
  **zero** PATH lookups — the zero-install claim stated directly rather than implied.

## Verification

| Gate | Result |
| --- | --- |
| `bun run typecheck` | 14/14 |
| `bun run lint` | 12/12 |
| `bun run fmt:check` | clean |
| `bun run test` — unit | **4072 pass / 0 fail** |
| `bun run test` — integration | **92 pass / 0 fail / 47 skip** |
| `bun run test` — providers | 264 pass |
| `bun run test` — docs | 71 pass |
| `bun run build` | 10/10 |
| `git diff --check` | clean |
| `verify:doc-paths` / `verify:doc-coverage` | pass |
| `check-codegen-freshness` | fresh |

The 47 integration skips are the repository's existing baseline, not exclusions
introduced here.

Static gates from the plan: no `arrayBuffer()` in the source cache, no `jpeg-js` in
`apps/cli/src`, no `decodeJpeg`/`decodeImageBytes`/`renderSixelFromBytes`/
`encodeNativePng`/`decodeToRgba`/`prepareKittyPayload`/`ensurePngBytes` anywhere in
source, and no TMDB `original` selection. The **built binary contains no
chafa/ImageMagick/hpjansson strings at all.**

## Honest verification gaps

- **No real-terminal check.** The iTerm2 and VS Code paths are unit-tested but I
  have neither terminal available, and renderer selection requires a TTY so it
  cannot be exercised through a pipe. Kitty, Sixel, Windows Terminal, tmux
  fallback, resize, and burst navigation are likewise unverified by hand.
- `APP_SHELL_SIXEL_MAX_COLORS` stays at 64. Raising it risks the stdout-blocking
  input lag already diagnosed once in this repo, and I could not measure the
  tradeoff here.

## Note for merge order

Touches `apps/cli/src/ui.ts` (`CapabilitySnapshot`), as does #33. Whichever lands
second needs a rebase; the conflict is mechanical (#33 adds `curl`, this removes
`chafa`/`magick`).

https://claude.ai/code/session_012N9Tm1cSoB5pvYjrThFvsz


KitsuneKode:
## Windows CI found a real pre-existing bug

The first run failed `Windows CLI parity` — **not** the Chocolatey 504 that hit
#32/#33 earlier today. I read the log rather than assuming, and it was
`test:unit`, on the new local-source tests:

```
(fail) fetchPosterSource — local bounds > reads a local sidecar within the limit
  Expected: 2048   Received: undefined
(fail) fetchPosterSource — local bounds > normalises a file:// URL to an absolute path identity
  Expected: "C:\Users\RUNNER~1\...\poster.jpg"   Received: undefined
```

`isLocalImagePath` only understood POSIX absolutes, so on Windows a downloaded
thumbnail at `C:\Users\...\poster.jpg` was classified as **remote**, handed to
`fetch()`, and failed silently — every local poster on Windows came back blank.

That logic is unchanged from before this PR. The bug was pre-existing and simply
had no test covering it; the new local-source suite is what exposed it.

Fixed in `a8861acf`: drive-absolute with either separator (`C:\x`, `C:/x`), UNC
shares (`\\server\share`), and Windows file URLs are recognised. Stripping
`file://` from `file:///C:/x` also drops the leading slash, since `/C:/x` is not a
path Windows can open. Also stops classifying a protocol-relative URL
(`//cdn/x.jpg`) as local — it satisfied the POSIX test, which was true of the
original logic too.

The new cases are asserted as **pure string logic so every platform runs them**,
rather than leaving Windows CI as the only place this is checked.

All checks now pass, `mergeStateStatus: CLEAN`. Unit count 4072 → 4085.

PR #119, opened 2026-08-23
fix(docs): stop committing provenance that changes on every regeneration

KitsuneKode:
Ends the merge-conflict class that hit **four consecutive PRs today** — #83, #85, #84, and #93 each conflicted on `apps/docs/lib/generated-metadata.json` and needed a manual regenerate to land.

## Cause

The file carried two fields that move on every regeneration regardless of content:

```json
"syncedAt": "2026-08-23T09:07:23.744Z",     // timestamp — changes every run
"cliSourceRevision": "c0d354c7",            // sha — changes every commit
```

So any two branches that ran `generate` produced different files and collided. The writer already knew these were noise — it excluded both when deciding *whether* to rewrite — it just still wrote them.

## Fix

They move to `lib/generated-provenance.json`, written on every generate and **gitignored**. The committed metadata is now stable: running `generate` twice in a row produces no diff.

`metadataIdentity` also loses its field-exclusion dance — with no churn in the file, identity is the whole payload, and `check-codegen-freshness.ts` keeps asking the identical question.

## Nothing user-visible is lost

`SyncedAt`, the docs sidebar banner, and `llms.txt` all read the same two values from the new module. The drift test still asserts the revision shape, now against provenance.

## One thing this caught

Deleting the sidecar and building — the state a fresh clone is in — failed with `Can't resolve './generated-provenance.json'`. `build` now runs `generate` first, matching what `dev` already did. Worth noting the fix deliberately avoids `fs` reads in `provenance.ts`: that is exactly the pattern #87 removed to stop Turbopack tracing the whole workspace.

## Verification

- `generate` twice → no diff on the committed file
- `check-codegen-freshness` → fresh
- docs build from a clean state → 43 static pages
- docs tests → 83 pass
- `lint`, `fmt`, `verify:doc-paths` → clean

https://claude.ai/code/session_01GssSecHjd7uJW5HFZMUVGu

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -96,3 +96,8 @@ artifacts/
 # public and this directory holds proprietary source that must never be pushed.
 # Ignored (not merely untracked) so a stray `git add -A` cannot stage it.
 docs/installer-reference/
+
+# Provenance for the generated docs metadata: regenerated on every build, and
+# deliberately not committed — syncedAt and cliSourceRevision change every run,
+# so committing them made any two docs branches conflict on the same file.
+apps/docs/lib/generated-provenance.json
--- a/apps/docs/app/llms.txt/route.ts
+++ b/apps/docs/app/llms.txt/route.ts
@@ -1,4 +1,5 @@
 import { codeMetadata } from "@/lib/code-metadata";
+import { provenance } from "@/lib/provenance";
 import { publishedReleaseNotesArtifacts, releasePath } from "@/lib/release-notes";
 import { docsSiteUrl } from "@/lib/site";
 import { source } from "@/lib/source";
@@ -12,8 +13,8 @@ export async function GET() {
     "# Kunai Docs",
     "",
     `@doc-version: ${codeMetadata.cliVersion}`,
-    `@cli-source-revision: ${codeMetadata.cliSourceRevision}`,
-    `@synced-at: ${codeMetadata.syncedAt}`,
+    `@cli-source-revision: ${provenance.cliSourceRevision}`,
+    `@synced-at: ${provenance.syncedAt}`,
     "",
     "> Terminal-first guides for Kunai playback, recovery, offline use, and diagnostics.",
     "",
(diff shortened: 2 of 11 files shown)


KitsuneKode:
Resolved the merge conflict with `main` (on `generated-metadata.json` — this PR was blocked by the exact problem it fixes), and added the automated gate the fix was missing.

## Conflict resolution

Merged `main`, took this branch's shape for the generated file, and regenerated. Verified after:

- `generate` twice → zero diff
- `check-codegen-freshness` → fresh
- docs tests 85 pass, full `bun run test` 14/14, typecheck 14/14

## New: `apps/docs/test/codegen-idempotence.test.ts`

This PR verified "generate twice → no diff" by hand. That is the property worth keeping, so it is now a test.

It deliberately compares **run N against run N+1**, not against the committed bytes — a merely stale checkout should fail `check-codegen-freshness` (its own gate), not be misreported here as nondeterminism. The two questions are different and neither implies the other:

| Gate | Question |
| --- | --- |
| `check-codegen-freshness` | is the committed file up to date with its inputs? |
| this test | is generating deterministic at all? |

Original bytes are restored in `afterAll`, so running the suite never rewrites a working tree that had legitimately stale output.

Watched failing before trusting it — putting `syncedAt` back into the committed payload gives:

```
generated-metadata.json changed between two consecutive generate runs.
Drifting keys:
  syncedAt: "…:01.553Z" -> "…:02.328Z"

A committed generated file must not depend on the clock, the git revision,
hostname, absolute paths, locale-dependent sorting, or iteration order.
Move run-specific values into lib/generated-provenance.json, which is gitignored.
```

Also asserts the sidecar stays gitignored, since that is the escape hatch the whole rule depends on.

## Audited the other generated files while here

No other committed artifact churns: `generated-mascot` and `generated-troubleshooting-faq` have no clock fields, and `generated-release-notes`' `publishedAt` is a fixed historical release date — content, not clock. `.gitattributes` already pins `eol=lf` repo-wide, so the Windows CRLF path is covered too.

PR #156, opened 2026-08-23
fix(history): stop -i/--id persisting its placeholder, and stop poor writes erasing external ids

### LINKED ISSUE #154: -i/--id persists the "TMDB <id>" placeholder to history, and a poor write erases a richer row's external ids
## Symptom

`kunai -i 438631 -t movie` (Dune) plays fine and the details panel shows the real
title, but History renders the row as **`TMDB 438631`** with a blank poster:

```
key:               movie:438631:none:none:none
title_id:          438631
title:             TMDB 438631      <-- placeholder, never upgraded
poster_url:        (null)
external_ids_json: (null)
provider_id:       videasy
```

### 1. The working `TitleInfo` is never enriched from the catalog detail

`-i/--id` synthesises a placeholder title — `name: "TMDB <id>"`, no poster, no
external ids (`apps/cli/src/app/bootstrap/bootstrap-intent.ts:72`, via
`directIdTitleName` in `apps/cli/src/domain/types.ts:65`).

When the catalog answers in `PlaybackPhase`, the local `title` is upgraded with
`upgradeTitleInfoStructure(title, catalogDetail.type)`
(`apps/cli/src/app/playback/PlaybackPhase.ts:748`) — which touches **only**
`type` and `episodeCount` (`apps/cli/src/domain/media/anilist-format.ts:50`).
The resolved name is patched **only** into session state
(`apps/cli/src/domain/session/SessionState.ts:492`).

Every persistence and broadcast surface reads the **local** `title`, not session
state, so all of these record the placeholder:

| Surface | Site |
| --- | --- |
| History finalize | `PlaybackPhase.ts:2451` (`title: title.name`, `posterUrl: title.posterUrl`) |
| Playback ledger start | `PlaybackPhase.ts:3891` |
| Discord presence | `services/presence/PresenceServiceImpl.ts:624` (`details: title.name`) |
| Share refs (`/share`) | `domain/share/share-ref-from-title-context.ts:21` |

Result: the panel says "Dune" while the DB — and anyone you send a share link to
— says "TMDB 438631".

The same class hits share/`--open` launches whose ref carries no title:
`resolve-share-target.ts:79` falls back to `name: "tmdb:438631"`, which
`isDirectIdTitleName` does not even recognise, so the session panel stays wrong
too.

### 2. `upsertProgress` lets a metadata-poor write destroy a richer row

`packages/storage/src/repositories/history.ts` upserts with
`title = excluded.title` and `external_ids_json = excluded.external_ids_json`,
while `poster_url` is `COALESCE`d — an inconsistency that is the tell.
Reproduced against a temp DB:

```
after good launch (search lane):
  title "Dune"        externalIds {tmdbId:"438631", imdbId:"tt1160419"}  poster OK
after `-i 438631 -t movie`:
  title "TMDB 438631" externalIds GONE                                   poster OK
```

Note `backfillTitleMetadata` in the same file already merges and never blanks —
so the replace-semantics in `upsertProgress` is an accident, not a policy.

## Blast radius — the worst version

Once `external_ids_json` is null and the title id is bare (`438631`, not
`tmdb:438631`), `resolveTmdbIdentity`
(`apps/cli/src/services/sync/sync-identity.ts:58`) returns `null`. The row
**silently stops mirroring to the user's TMDB account** — no error, no
diagnostic. Watch a title by search (syncs correctly), later open the same title
with `-i`, and the good name is overwritten, the ids are erased, and tracker sync
goes quiet. That is the house "silent no-op" failure mode, with real user data
lost rather than merely mis-rendered.

Secondary: for `-i <id> -t series`, release reconciliation and "new episode"
notifications key off the same row, so notifications read `TMDB <id> — new
episode`.

The existing self-healer cannot recover any of it: `HistoryMetadataHealer`
searches by the row's **title text** (`create-history-metadata-resolver.ts`), so
it queries the catalog for the literal string "TMDB 438631", and
`ResolvedHistoryMetadata` has no `title` field to write back even on a hit.

## Fix

1. **Enrich at the seam.** One pure domain function that adopts `name` (only
   while the current name is a placeholder for that id), `posterUrl`, merged
   `externalIds`, `year`, and `episodeCount` from a `TitleDetail`. Call it from
   `PlaybackPhase` where `catalogDetail` arrives, and from the `SET_TITLE_DETAIL`
   reducer, so state and the persisted row cannot disagree. Generalise
   `isDirectIdTitleName` to also recognise the `ns:id` share fallback.
2. **Make the storage write non-destructive.** In `upsertProgress`, keep the
   richer title when the incoming one is a placeholder for that id, and merge
   external ids instead of replacing them. This closes the class, not just this
   lane.
3. **Repair rows already written.** Teach heal-target selection to flag a
   placeholder title, resolve it by **id** through the catalog detail service
   (exact) rather than by searching the placeholder text (fuzzy, and here
   guaranteed to miss), and let the backfill write a corrected title.

## Repro

```sh
kunai -i 438631 -t movie   # play a few seconds, quit
# History now shows "TMDB 438631" with no poster
```

KitsuneKode:
Closes #154.

## What was wrong

`kunai -i 438631 -t movie` played Dune, and the details panel said so — but
History recorded the row as `TMDB 438631`, with no poster and no external ids.

Two defects compounded.

### 1. The working `TitleInfo` was never enriched from the catalog

`-i/--id` synthesises a placeholder title — `name: "TMDB <id>"` and nothing
else. When the catalog answered, `PlaybackPhase` upgraded only its **type**
(`upgradeTitleInfoStructure` touches `type`/`episodeCount`), and the resolved
name was patched only into session state.

Everything that persists or broadcasts reads the **local** `title`:

| Surface | Site |
| --- | --- |
| History finalize | `PlaybackPhase.ts` |
| Playback ledger start | `PlaybackPhase.ts` |
| Discord presence | `PresenceServiceImpl.ts` (`details: title.name`) |
| Share refs (`/share`) | `share-ref-from-title-context.ts` |

So the panel said "Dune" while the database — and anyone the share link reached
— said "TMDB 438631". The same class hit `--open` on a share ref carrying no
title, which falls back to naming a title after its own id (`tmdb:438631`).

### 2. A metadata-poor write destroyed a richer row

`upsertProgress` replaced `title` and `external_ids_json` outright, while
`poster_url` was `COALESCE`d — the inconsistency was the tell. Reproduced:

```
after a search-lane launch:  title "Dune"        ids {tmdbId, imdbId}  poster OK
after `-i 438631 -t movie`:  title "TMDB 438631" ids GONE              poster OK
```

**Why that is more than cosmetic:** with the ids gone and a bare title id,
`resolveTmdbIdentity` returns `null`, so the row **silently stops mirroring to
the user's TMDB account** — no error, no diagnostic. Watch a title by search
(syncs fine), later open it with `-i`, and sync goes quiet. The end-to-end test
pins exactly this, before and after.

## The fix

**1 — Enrich at the seam.** New `domain/catalog/apply-title-detail.ts` folds the
whole catalog answer into the working title: name (only while it is a
placeholder for that id), poster, merged external ids, year, episode count.
Both `PlaybackPhase` and the `SET_TITLE_DETAIL` reducer go through it, so the
rendered title and the persisted one cannot disagree. Adoption is upgrade-only:
a searched name, a provider poster, and ids the title already carries all win.

**2 — Make the storage write additive on identity.** Progress fields still come
from the newest write; identity metadata is protected. A placeholder title never
replaces a stored real name, and external ids merge with stored ids winning per
key. This closes the class, not just this lane — any poor-metadata lane is now
harmless.

**3 — Repair rows already written.** Heal-target selection flags a placeholder
title, and the resolver resolves those by **id** through the catalog rather than
by searching the placeholder text — which could only miss, or match an unrelated
row that happens to share the number. A placeholder over an opaque
provider-native id is deliberately **not** flagged: nothing could resolve it, and
a missed repair is recoverable where a wrong one is not.

The placeholder vocabulary (`directIdTitleName` / `isPlaceholderTitleName`)
moves to `@kunai/core`, because the CLI mints it and storage has to refuse to
persist it — they must agree on one definition.

## Tests

- `apply-title-detail.test.ts` — 12 cases: adoption, refusal to overwrite real
  values, the `ns:id` share fallback, structure upgrade, identity return.
- `history-metadata-preservation.test.ts` (storage) — placeholder never
  overwrites, real title still heals forward, an ordinary rename still applies,
  ids merge, progress still advances, backfill repairs only placeholder rows.
- `heal-placeholder-title.test.ts` — selection, id-based resolution, refusal to
  text-search a placeholder, refusal of a catalog answer that is itself a
  placeholder, opaque ids left alone.
- `heal-placeholder-row-end-to-end.test.ts` — the exact corrupted row, over a
  real SQLite database, healed by the real repository/selector/resolver/healer,
  asserting the tracker identity goes from `null` to resolved and that a second
  pass is a no-op.

## Verification

`bun run typecheck`, `bun run lint`, `bun run fmt:check`, `bun run test`
(4639 unit + 145 integration + all packages, 0 fail), `bun run build`,
`verify:doc-paths`, `verify:doc-coverage`, `verify:doc-frontmatter` — all green.
`.docs/architecture.md` documents the new write contract.

https://claude.ai/code/session_01KFHYvhwf29JhDFNbaJBgCx


KitsuneKode:
Pushed a second commit closing two more instances of the same root cause, found by walking the other consumers of a metadata-poor `TitleInfo`.

### `--download` never reaches `PlaybackPhase`

`maybeRunDownloadMode` handed the `-i/--id` placeholder straight to `DownloadOnlyPhase` → `commitDownloadIntent`. Nothing on that path fetches catalog detail — `prepareConfirmedTitle` only does anime provider mapping. So `kunai -i 438631 -t movie --download` wrote `TMDB 438631` into:

- `download_jobs.title_name`
- the offline-library asset title
- **the file on disk** — `resolveDownloadOutputPath` builds the folder and file stem from the title

That is worse than the history row it mirrors: the startup healer in the first commit can repair a database, but nothing can rename a file the user already has.

`app/bootstrap/resolve-placeholder-title.ts` now resolves it up front, reusing `applyCatalogDetailToTitle`. Only a placeholder over a catalog-addressable id costs a lookup, and any failure returns the title untouched — a download with an ugly name beats a download that did not happen.

### `list_items` and `followed_titles` had the same clobber as history

```
lists.ts             title = excluded.title,  external_ids_json = excluded.external_ids_json
followed-titles.ts   title = excluded.title
```

Re-adding a title to a Watchlist, or muting one, rewrote both fields from whatever the caller happened to hold. Both now follow the history rule: a placeholder never replaces a stored real name, ids merge with stored ids winning per key, and an ordinary rename still applies. Muting still changes the preference — it just no longer renames the title.

### Checked and deliberately left alone

- `queue.ts` — no title/ids clobber.
- `notifications.ts`, `release-progress-cache.ts` — replace `title`, but those are per-record and cache rows respectively.
- `-i` with `-a`, and `-i` without `-t` — rejected before anything is persisted.
- share / `--open` playback launches — covered by the first commit, since `fetchTitleDetail` echoes the same id back.

11 new tests; full gate green again (typecheck, lint, fmt:check, 4644 unit + 145 integration + all packages, build, doc verifiers).

KitsuneKode:
@coderabbitai review

PR #197, opened 2026-08-24
fix(providers): key every resolve cache on the whole preference set

### LINKED ISSUE #187: Provider cache keyParts disagree, so preference switches serve stale streams
Manifest `keyParts` are load-bearing — `stream-resolve-cache.ts` reads them verbatim — and they disagree across production providers, so a preference switch can reuse a cache entry that encodes a different preference until TTL.

- **videasy**: missing `audio` entirely (`manifest.ts:29-41`). A dub↔sub switch reuses the same key. It is the only production manifest missing it; allmanga/anidb/miruro all have it.
- **vidlink**: missing `quality`, `startup`, `source`, `stream`, `audio` — narrower than its own behavior, so a cached `fast` result can serve a `quality-first` request.
- **rivestream**: missing `audio` while audio preference now drives cycle ordering (see #185).
- **miruro**: missing `subtitle`, so an en→es switch can serve stale-subtitle-fact streams.

Deeper variant in videasy route policy: `cineby-vyse` (English filter) and `cineby-fade` (Hindi filter) share the endpoint `wings-hdmovie`, producing identical route cache keys (`direct.ts:900-915`). Whichever flavor resolves first owns the entry and the other language can get wrong-language streams.

Suggested fix is a manifest conformance test that makes the drift impossible rather than fixing five manifests by hand: assert required `keyParts` per declared capability (`audio-modes` ⇒ `audio`; subtitle resolve ⇒ `subtitle`), plus a cross-provider matrix pinning `keyParts` equality where behavior is equal. Today five manifests encode five private opinions.

KitsuneKode:
Stacked on #185. Closes #187.

Manifest `keyParts` are read verbatim to build the CLI stream-resolve cache key. Several production manifests omitted request preferences, and YouTube used an unsupported literal `video-id` token, so different videos produced the same cache key.

This change:

- adds the complete preference identity to Vidlink, Videasy, Rivestream, Miruro, and AniDB manifests;
- keys YouTube on the real title/video identity plus the complete preference set;
- moves cache-key conformance to `bootstrap-providers.test.ts`, where it derives coverage from `loadProductionProviderModules()` instead of maintaining a second hand-written provider list;
- proves two distinct YouTube video ids cannot share a stream-resolve key;
- proves Vyse and Fade remain distinct through their source ids even though both use `wings-hdmovie`.

The Vyse/Fade route-policy concern from #187 does not require a second route-specific lookup cache. Videasy's selected-route policy is result provenance and TTL metadata; the CLI read/write/invalidation key is manifest-driven and now contains both source and audio request identity. The provider guide records that boundary explicitly.

Verification on commit `42ef8bd2`:

- pre-push: 23/23 Turbo tasks passed, including build;
- CLI unit: 4,715 passed;
- CLI integration: 147 passed, 24 opt-in skips;
- focused cache/provider contracts: 38 passed;
- typecheck: 14/14 tasks passed;
- lint: exit 0, no errors introduced by this slice;
- formatting and doc-path verification passed.


KitsuneKode:
@coderabbitai review

PR #198, opened 2026-08-24
fix(providers): bound the miruro and rivestream module caches

KitsuneKode:
Stacked on #197 (→ #185). Closes #190.

All instances of the growth class already fixed for the Videasy Wings transports:

- **Miruro** `episodeCache`/`sourceCache` were `TTLCache`s with no ceiling → one entry per title/episode probed, never freed. Both now carry `maxEntries`.
- **Miruro** had no test seam, so the smoke's `KITSUNE_CLEAR_CACHE=1` cleared the CLI store but not the module caches. `clearMiruroCachesForTest` now exists and the smoke calls it.
- **Rivestream** `secretKeyCache` was an unbounded `Map` that also cached the `generateSecretKey` failure sentinels, so one transient error pinned a broken key until process exit. Now a bounded `TTLCache`, and sentinels are never cached.

`maxEntries` eviction is already covered in `provider-cache.test.ts`; new tests guard the wiring and the clear seam.

Gate: typecheck 14/14, tests 23/23 forced, zero lint errors.

https://claude.ai/code/session_01BkWab6pnVL5vMsVGRv9TVp

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Miruro and Rivestream playback reliability by managing cached provider data within defined limits.
  * Prevented temporary Rivestream secret-generation failures from being reused.
  * Added safer cache clearing for Miruro-related playback data.

* **Tests**
  * Added coverage verifying cache limits, clearing behavior, and successful Rivestream playback candidate creation.

PR #199, opened 2026-08-24
fix(providers): wire the miruro subtitle boost, drop dead videasy constants

### LINKED ISSUE #191: Silent no-ops across providers: declared capabilities with no reader
Three instances of the documented house failure mode (`AGENTS.md`: "flags parsed and dropped, settings persisted and ignored, capabilities declared and unread"), which `contract-conformance.test.ts` is supposed to gate.

- **miruro** `preferredSubtitleDelivery` applies a −5000 priority boost (`direct.ts:573`) but is never passed by the production resolve path (`direct.ts:1720`). The parameter exists, is honoured, and is never supplied.
- **videasy** exports `WINGS_DB_BASE` / `SPEEDRACE_DB_BASE` to no reader.
- **videasy** relay `upstreamHosts` lists hosts that no code path contacts.

Also misleading rather than inert: videasy Phase B background probe rows hardcode host `api.videasy.to`, while the wings endpoints actually live on `speedracelight.com` — every lazy-inventory row shows the wrong host. And plaintext upstream error bodies are fed to the PRNG decryptor and surface as fake "tampered payload" crypto errors, which sends debugging in the wrong direction.

Worth checking why the conformance gate does not catch these — the gate may be narrower than the rule it is meant to enforce.

KitsuneKode:
Stacked on #198 (→ #197 → #185). Closes #191.

Two declarations with no reader (the documented house failure mode):

- **Miruro** gives hard-sub `sub` servers a −5000 priority boost, but the resolve path never passed `preferredSubtitleDelivery` into the builder, so it was dead. The adapter already sets it to `hardcoded` for anime; now threaded through. A test proves the boost reorders sub ahead of dub.
- **Videasy** exported `WINGS_DB_BASE`/`SPEEDRACE_DB_BASE`, read by nothing. Removed (`VIDEASY_DB_BASE` beside them has 4 callers, kept).

Not changed: the audit also flagged videasy's relay `upstreamHosts` as unread, but `packages/relay/src/registry.ts` matches request hosts against it and the list is accurate — it is wired.

### DIFF
--- a/packages/providers/src/miruro/direct.ts
+++ b/packages/providers/src/miruro/direct.ts
@@ -1753,6 +1753,14 @@ export const miruroProviderModule: CoreProviderModule = {
         targetAudio,
         fallbackAudio,
         preferredSourceId: input.preferredSourceId,
+        // The builder gives hard-sub sub servers a priority boost, but this was
+        // never supplied, so the boost was dead. The adapter sets this to
+        // "hardcoded" for anime; `external` has no Miruro equivalent, so it maps
+        // to `unknown` and simply does not trigger the boost.
+        preferredSubtitleDelivery:
+          input.preferredSubtitleDelivery === "external"
+            ? "unknown"
+            : input.preferredSubtitleDelivery,
       });
       const sourceInventorySeeds = buildMiruroSourceInventoryCandidates(
         cycleCandidates,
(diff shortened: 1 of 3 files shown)


KitsuneKode:
@coderabbitai review

Issue #195, opened 2026-08-24
Miruro resolve is ~10s on the success path, with no per-stage trace


KitsuneKode:
Measured live 2026-08-24 (One Piece ep1159, `balanced`): `resolveDurationMs: 10324`, `failureCodes: []`, stream reachable. Earlier audits reported ~24s.

Two separate things, both worth correcting in the record:

1. The ~24s figure is the **engine attempt budget**, not a Miruro defect — `balanced` is a 12s attempt timeout × 2 attempts. #185 clamped per-candidate timeouts below the attempt budget so a hung mirror can no longer consume the whole attempt, which removes the pathological case.
2. The remaining ~10s is the **success path** — no candidate failed on that run, so the candidate bound never fired. The time is in sequential Cloudflare-gated pipe calls (providers → episodes → sources), not candidate churn. #185 did not improve this and should not be credited with it.

Blocked on diagnosis: the `--debug` log carries no per-stage provider trace timings, so the 10s cannot currently be attributed to a stage. Add stage timings to the resolve trace first, then optimise against evidence.

Explicitly **not** recommending parallelising the pipe calls the way rivestream prefetches candidates: rivestream is not WAF-gated the same way, and burst concurrency against Cloudflare risks provoking the blocking this provider already fights.

KitsuneKode:
Related work: #205 covers persistent caching and #267 covers endpoint health semantics. Keep this ticket focused on Miruro stage-level latency/trace evidence so the performance work can be sequenced without duplicating the cache or health changes.

PR #200, opened 2026-08-24
feat(miruro): attribute resolve latency with per-stage trace timings

KitsuneKode:
Stacked on #199 (→ #198 → #197 → #185). Refs #195 (does not close it — see below).

The ~10s success-path latency could not be diagnosed because no stage was timed. Resolve now emits `durationMs` trace events for the two stages that spend the time: the episode-catalog fetch (`source:miruro:episodes`) and the source-resolution cycle (`source:miruro:source-cycle`).

The episodes fetch is timed **inclusive of failure** — a Cloudflare block throws there, which is the slow case most worth attributing.

**Live evidence:** of 8061ms total, **6898ms was the episodes fetch**. So the bottleneck is the gated pipe call, not source resolution. Any optimization should target that.

Instrumentation only — no resolve behavior changes. I kept #195 open deliberately: this gives it the evidence, but the actual latency reduction is separate work (and I would not parallelize the WAF-gated pipe calls, which risks provoking the blocking).

### DIFF
--- a/packages/providers/src/miruro/direct.ts
+++ b/packages/providers/src/miruro/direct.ts
@@ -1724,8 +1724,29 @@ export const miruroProviderModule: CoreProviderModule = {
       startupPriority: input.startupPriority,
     });
 
+    // Time the episodes fetch inclusive of failure: a Cloudflare block throws
+    // here, and that is the case whose latency most needs attributing.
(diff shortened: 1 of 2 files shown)


KitsuneKode:
@coderabbitai review

Issue #193, opened 2026-08-24
Anime audio fallback semantics contradict between miruro and anidb


KitsuneKode:
Same product, two semantics for the same user intent.

- **miruro** falls back dub→sub silently when the requested dub server is down. The user asked for English audio, gets Japanese, and is told nothing.
- **anidb** has an explicit documented contract that a missing requested language must **not** fall back.

Both ship in the anime lane, so which one the user gets depends on which provider resolved.

Decide the product rule, then make both obey it, and emit a typed `resolved-via-fallback-audio` trace when a downgrade happens so the shell can surface it instead of silently changing what the user hears.

Related: audio tracks are second-class outside the anime trio. `videasy`, `vidlink` and `rivestream` emit no `inventory:audio-modes`; videasy ignores `preferredAudioLanguage` unless a source id is pinned. #185 made rivestream *route* by requested language rather than only labelling streams with it, but it still stamps the requested language onto streams with no provider evidence (confidence 0.65) — the evidence problem is unfixed.

Any provider that can observe mode evidence should be required to emit `inventory:audio-modes`.

PR #201, opened 2026-08-24
feat(miruro): make the silent dub<->sub audio downgrade visible

KitsuneKode:
Stacked on #200 (→ #199 → #198 → #197 → #185). Refs #193.

Miruro walks its fallback audio when the requested one has no working server, so a dub request quietly resolves a sub with nothing telling the user. Resolve now emits a typed `audio:fallback` trace event (`requested` vs `resolved`) whenever the resolved presentation differs from the request, via a unit-tested pure predicate `isMiruroAudioFallback`.

**AniDB is left as is** — it already fails closed on the requested mode (`No AniDB streams … (dub)`), which is the documented no-fallback contract the two were said to disagree on. So only Miruro needed the signal, and the "two semantics" is resolved by making Miruro's downgrade visible rather than removing it.

The new `audio:fallback` event type is additive — nothing exhaustively switches on the union (consumers `.find` specific types).

Left #193 open: this surfaces the downgrade in the trace; wiring the shell to actually show it to the user is a follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added audio fallback notifications when the requested dub or sub mode is unavailable and another mode is selected.
  * The interface can now indicate when the resolved audio language differs from the requested preference.

* **Bug Fixes**
  * Preserved strict audio-mode behavior for providers that do not support fallback, preventing silent mode changes.


KitsuneKode:
@coderabbitai review

Issue #120, opened 2026-08-23
SakuraLoader cannot render during a bootstrap search on any surface but idle


KitsuneKode:
Fixed narrowly in #93: `RootIdleShell` now renders `SakuraLoader` while a bootstrap `-S` search resolves. Before that the root chrome said "searching" over a welcome screen with no sign of work in flight.

The underlying shape is still there. A bootstrap search runs *before* `openBrowseShell` is called, so the browse shell — which owns the real loading surface at `browse-shell.tsx:1917` — is not mounted yet. The idle shell now duplicates a slice of that responsibility.

This is the "single mounted content tree" roadmap item (`.plans/phase-1.8-single-mounted-content-tree.md`): mount the shell first and let the search stream into it, and the duplication disappears along with a class of similar gaps.

Not urgent — the visible symptom is fixed. Filed so the structural cause is not forgotten once the symptom stops hurting.

KitsuneKode:
Closing as superseded by the persistent-shell roadmap. The user-visible bootstrap loading gap was fixed; the remaining concern is architectural duplication already tracked by the shell migration work.

PR #202, opened 2026-08-24
fix(shell): show the Sakura loader while a selected title resolves

KitsuneKode:
Standalone (based on `main`, not the provider stack).

**The "why":** selecting a series/movie dispatches `SELECT_TITLE`, which sets `view: "details"` with `playbackStatus: "idle"`. For the whole window before playback resolves, the idle root surface (`RootIdleShell`) rendered its static session view — the blank screen in the report. `RootIdleShell` only rendered the Sakura loader for a bootstrap search (`searchState === "loading"`), never for this resolve window. That's why some loading instances show the loader and this one didn't.

**The fix:** show the shared `SakuraLoader` when a title is selected and awaiting playback. `view === "details"` is a precise trigger — every playback path, movie or series, later dispatches `SELECT_EPISODE`, which flips the view to `"playback"`, so a resting or paused session never sits at `"details"`. The loader therefore cannot appear on the resume screen (a real risk I verified against).

Rendered frame:
```
       Preparing The Rookie…
✿ ❀ ✿  Resolving sources · esc to cancel
```

Tests cover all four idle states (preparing / resting-session / bootstrap-search / welcome); the resting-session guard fails without the fix.

Structurally this is the same gap as #120 (surfaces owning their loader not being mounted during a transition), fixed one step later in the flow.

### DIFF
--- a/apps/cli/src/app-shell/root-status-shells.tsx
+++ b/apps/cli/src/app-shell/root-status-shells.tsx
@@ -40,6 +40,25 @@ export function RootIdleShell({ state }: { state: SessionState }) {
       </Box>
     );
   }
+
+  // A title was selected but playback has not started yet. `SELECT_TITLE` sets
+  // `view: "details"` with `playbackStatus: "idle"`, and every playback path —
+  // movie or series — later dispatches `SELECT_EPISODE`, which flips the view to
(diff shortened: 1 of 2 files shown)


KitsuneKode:
@coderabbitai review

Issue #179, opened 2026-08-24
Move tracker and provider secrets behind an OS credential vault


KitsuneKode:
## Verified current behavior

SyncTokenStore writes AniList and TMDB credentials to an owner-only JSON file, and videasySessionToken remains in config.json. File permissions reduce exposure but do not provide OS-backed encryption at rest.

## Required design

Introduce one credential-vault port owned by the persistence layer, with implementations for macOS Keychain, Windows user-scoped protection or Credential Manager, and Linux Secret Service when available. Define and test the headless Linux fallback explicitly rather than making first run fail when no desktop keyring exists.

## Migration and rollback

- migrate AniList, TMDB, and Videasy secrets without logging or duplicating values
- delete plaintext only after a verified vault write
- make migration restart-safe and idempotent
- preserve disconnect, reset, export, and uninstall semantics
- never expose secrets through settings metadata, diagnostics, support bundles, or subprocess argv
- support rollback to the previous application version without credential loss

## Acceptance

- fake vault contract tests cover read, write, delete, locked, unavailable, and partial migration paths
- native Windows and macOS CI exercise their real OS protection APIs
- Linux covers Secret Service plus headless fallback policy
- existing plaintext fixtures migrate once and are removed only after verification
- no new native build dependency or postinstall script is introduced

This is security hardening, but it should not be folded into the archive and attestation release train without the native migration evidence above.

PR #211, opened 2026-08-24
fix(storage): make secret writes owner-only on Windows and durable everywhere

KitsuneKode:
Refs #179. This is the proportionate slice of the credential-vault work — the part that measurably improves security without coupling a keychain migration to the updater and rollback paths this release is already rewriting.

`writeAtomicSecretJson` is the single path that persists AniList and TMDB tokens (`sync-tokens.json`) and the Videasy session token (`config.json`). It had two gaps.

### 1. Windows got no restriction at all

```ts
if (posixMode !== undefined && process.platform !== "win32") {
  await chmod(tmp, posixMode);
}
```

NTFS ignores POSIX mode bits, so on Windows the token file simply kept whatever `%APPDATA%` inherited. That is user-scoped by default — which is what most CLIs lean on — but it was an inherited default this process never established.

Now an inheritance-free, user-only ACL is applied via `icacls` **before** the rename. It is best-effort by construction: no `icacls`, or a policy that refuses, must not fail the write. If the new ACL would lock this process out, inheritance is restored rather than leaving an unreadable secret behind.

### 2. Nothing was ever flushed — on any platform

Temp-file-plus-rename stops readers seeing a half-written file, but on its own it is not durable. The rename can reach the filesystem journal while the data blocks are still in the page cache, so a power loss leaves a **correctly named, zero-length config**. `FileStorage.ts` already has a corrupt-config recovery path, which suggests this has been hit.

- the file is flushed before the rename makes it visible
- the directory entry is flushed after the rename (POSIX; Windows has no directory handle to sync)
- the mode is now set at creation, closing the window where a secret briefly existed with the process umask's permissions

### Coverage

The existing tests only exercised `atomicMove`. Added: secret writes are `0o600`; a rewrite over a `0o666` file narrows it; ordinary writes are not silently narrowed; `writeAtomicBytes` round-trips `Uint8Array` / `ArrayBuffer` / `Blob`; a directory flush never fails a landed write; ACL hardening is skipped rather than guessed when the user is unknown.

Gate: typecheck 14/14, lint 0 warnings, **5709 pass / 0 fail**, cache-forced.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file-save durability to help prevent data loss after crashes or interruptions.
  * Strengthened protection for secret files across operating systems.
  * Ensured text, JSON, and binary file writes consistently preserve their contents.
* **Tests**
  * Expanded coverage for file durability, permissions, secret-file protection, and binary data handling.

### DIFF
--- a/apps/cli/src/infra/fs/atomic-write.ts
+++ b/apps/cli/src/infra/fs/atomic-write.ts
@@ -1,4 +1,5 @@
-import { chmod, lstat, mkdir, rename, unlink } from "node:fs/promises";
+import { constants as fsConstants } from "node:fs";
+import { access, chmod, lstat, mkdir, open, rename, unlink } from "node:fs/promises";
 import { basename, dirname, join } from "node:path";
 
 function tempPath(targetPath: string): string {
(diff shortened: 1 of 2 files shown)


KitsuneKode:
@coderabbitai review

KitsuneKode:
@coderabbitai review

PR #212, opened 2026-08-24
docs(release): add the runbook for withdrawing a released version

KitsuneKode:
`.docs/release-reliability-gate.md` covers install, update, rollback and uninstall. It had no section on what to do when a bad release ships anyway — the one runbook you cannot write calmly during an incident.

**No code was needed.** Both update channels already resolve a server-side mutable pointer, so withdrawal is a reversible metadata edit:

| Channel | Pointer | Read by |
| --- | --- | --- |
| `binary` | GitHub `releases/latest` → `tag_name` | `latest-version.ts` |
| `npm-global` / `bun-global` | `registry.npmjs.org/@kitsunekode%2fkunai/latest` | `UpdateService.ts` |

Two properties make this work, both verified in the source rather than assumed:

- `parseVersionFromTag` **rejects prerelease tags**, so `gh release edit v0.3.0 --prerelease` drops the release out of `releases/latest` and clients resolve the previous stable one.
- Moving the npm `latest` dist-tag does the same for npm/bun without `npm unpublish`, which is 72-hour limited and hard-breaks pinned users.

The runbook also records two things that are easy to get wrong under pressure:

- **Do not delete the release or its assets.** Anyone pinned to that version still needs them to roll back, and deletion is not reversible.
- **Rehearse a real `kunai rollback`, not just `--dry-run`** — it is the least-exercised path that matters most, and it runs precisely when everything else has already failed.

Gates: `verify:doc-paths` (48 docs, 1962 source files, all paths resolve), `verify:doc-coverage`, `fmt:check`.

### DIFF
--- a/.docs/release-reliability-gate.md
+++ b/.docs/release-reliability-gate.md
@@ -301,3 +301,75 @@ bun run test:live:youtube
 ```
 
 Expected: `ok: true`, `streamResolved: true`, `streamHost` contains `youtube.com`. When yt-dlp is intentionally absent on the runner, `skipped: true` is acceptable.
+
+## Withdrawing a Released Version
+
+Every gate above exists to stop a bad release shipping. This section is for when
+one ships anyway. Rehearse it before tagging, not during an incident.
+
+Nothing here needs a code change. Both update channels resolve a *server-side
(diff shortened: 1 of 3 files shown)


KitsuneKode:
@coderabbitai review

PR #213, opened 2026-08-24
test(live): surface VidLink stream reachability, retire the vidking alias

KitsuneKode:
Two small gaps on top of the VidLink smoke this chain already added (`vidlink-inception.smoke.ts`, from *"stop transient probe failures condemning live sources"*).

Rebased onto the provider chain tip so it extends that smoke rather than duplicating it.

## 1. Reachability is measured and reported

Resolving a URL and serving bytes are separate claims, and for VidLink they had diverged — the CDN rate-limits CLI probes, so on `main` every resolve was discarded at the reachability gate.

**This chain repairs that, and the smoke now demonstrates it.** Three consecutive runs on the chain tip:

| Run | streamResolved | probe status | streamReachable | resolve |
| --- | --- | --- | --- | --- |
| 1 | true | reachable | true | 2008 ms |
| 2 | true | reachable | true | 1311 ms |
| 3 | true | reachable | true | 1375 ms |

The same smoke on `main` answers HTTP 429. So this is evidence the chain's transient-probe fix works, not just an assertion that it should.

**Reported, not enforced.** The exit code still turns on resolution alone. A single probe is exactly the transient signal this chain decided must not condemn a live source, so failing the smoke on it would re-introduce the behaviour that was just removed. The payload carries the truth; the gate keeps the chain's policy. Worth a reviewer's opinion on whether that's the right split.

## 2. Retires the dead `test:live:vidking` alias

```jsonc
"test:live:vidking": "bun run test:live:videasy"   // names a provider not in the tree; runs a different one
```

There is no `vidking` module. The one script that looked like vidking coverage exercised videasy. Remaining references are confined to `.archive/` (no authority) and `CHANGELOG.md` (history), both left alone.

## Gates

`typecheck` clean, `lint` 0 warnings across all 9 packages, `fmt`, plus the `guard` hook on `apps/cli/package.json`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved live stream checks by reporting whether resolved streams are reachable, including probe details in test results.
  - Increased the documentation test timeout to 20 seconds to reduce failures during longer-running test scenarios.

- **Documentation**
  - Clarified timeout guidance, including platform-specific testing considerations and when longer timeouts apply.

- **Chores**
  - Removed the obsolete VidLink live-test command.

### DIFF
--- a/apps/cli/package.json
+++ b/apps/cli/package.json
@@ -76,7 +76,6 @@
     "test:live:offline-beta": "bun -e \"await import('./test/live/offline-beta.smoke.ts')\"",
     "test:live:videasy": "bun test/live/videasy-bloodhounds.smoke.ts",
     "test:live:videasy:suite": "KUNAI_VIDEASY_LIVE_SUITE=1 bun test/live/videasy-bloodhounds.smoke.ts",
-    "test:live:vidking": "bun run test:live:videasy",
     "test:vhs:setup": "vhs test/vhs/setup.tape",
     "test:vhs:offline": "vhs test/vhs/offline-library.tape",
     "test:vhs:palette": "vhs test/vhs/command-palette.tape",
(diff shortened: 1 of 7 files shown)


KitsuneKode:
@coderabbitai full review

PR #214, opened 2026-08-24
fix(diagnostics): surface measured stage durations in summarised traces

KitsuneKode:
Closes the "instrumentation with no reader" gap on PR #200's per-stage timings.

#200 added `durationMs` to Miruro's stage events. Nothing downstream could read it, so the instrumentation could not answer the question it was added for — attributing Miruro's latency still meant re-reading raw trace events by hand.

## Three independent drops, each sufficient on its own

1. **`summarizeProviderTraceEvent` never mapped it.** It carried `failureClass`, `serverId` and `stage` across from the event, but not `durationMs`. (It is a top-level field on `ProviderTraceEvent`, not an attribute.)
2. **The source-event filter excluded `source:success`.** Timed stages report their duration on the event that *ends* them — and a stage that ends well ends on success. Filtering to `source:start` / `source:failed` therefore dropped every healthy stage's timing, which is most of the latency on a successful resolve.
3. **The smoke payload only attached the summary on failure.** `providerSmokeError` was the sole path that called `summarizeProviderTraceEvents`, so a resolve that *worked* reported one total duration and nothing about where it went — exactly backwards, since the latency worth attributing lives in successful resolves.

A measured stage is now canonical in its own right. Stage timings carry no attempt, server, or failure class, so keying "canonical" on those alone discarded them whenever any per-attempt event existed alongside.

## Result — Miruro now attributes its own latency

```
total resolveDurationMs: 9017
  source:miruro:episodes        3139 ms
  source:miruro:source-cycle    5874 ms
  source:miruro:pipe:kiwi:sub        candidate-timeout
  source:miruro:pipe:pewe:sub        (success)
```

3139 + 5874 ≈ the full 9017 ms. The two stages plus the dead `kiwi` candidate inside the cycle now explain the whole resolve, from the payload alone — which means CI can finally regress-test the latency #200 was written to attribute.

## Scope

The fix is in `summarizeProviderTraceEvents`, so it reaches **every** consumer of a summarised trace — the live smokes and the diagnostics panel alike — rather than being patched into one smoke.

## Gates

typecheck 14/14, **5814 pass / 0 fail** (cache-forced), lint identical to the chain baseline (2 + 7 pre-existing warnings, none introduced). New unit test pins the real Miruro event shape so all three drops stay closed.

### DIFF
--- a/apps/cli/test/live/provider-smoke.ts
+++ b/apps/cli/test/live/provider-smoke.ts
@@ -114,8 +114,17 @@ export function buildProviderSmokePayload({
   readonly resolveDurationMs?: number | null;
 }): ProviderSmokePayload {
   const engine = stream?.providerResolveResult?.trace.runtime ?? null;
+  // Summarise the trace on success too. It used to be attached only by
+  // `providerSmokeError`, so a resolve that worked reported one total duration
+  // and nothing about where the time went -- exactly backwards, since the
+  // latency worth attributing lives in successful resolves. On failure
+  // `providerSmokeError` is spread after this and still wins.
+  const traceSummary = summarizeProviderTraceEvents(stream?.providerResolveResult?.trace.events);
   return {
     ok: Boolean(stream?.url),
+    traceEventCount: traceSummary.eventCount,
+    lastTraceEvent: traceSummary.lastEvent,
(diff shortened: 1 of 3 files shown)


KitsuneKode:
@coderabbitai full review

PR #215, opened 2026-08-24
test(live): probe stream reachability in the rivestream, allanime and youtube smokes

KitsuneKode:
Stacked on #214.

Only `videasy` and `miruro` measured whether a resolved URL actually serves bytes. The other three asserted **resolution alone**, so a route that resolves a dead URL passed the release matrix — the exact false green VidLink proved is possible, where a well-formed URL is answered `429` by the CDN.

## What changed

`rivestream`, `allanime` and `youtube` now probe and report `streamProbe` / `streamReachable`, and a **measured** failure fails the smoke.

The guard is deliberately `streamReachable === false`, not `!streamReachable`:

- `false` → positive evidence the stream will not play → fail
- `null` → the probe never ran because nothing resolved → the existing resolution check already covers it, so it must not double-fail

## Verified live on the chain tip

| Smoke | streamResolved | probe | streamReachable |
| --- | --- | --- | --- |
| rivestream | true | `reachable` | **true** |
| youtube | true | `reachable` | **true** |
| allanime | false | — | **null** (captcha gate leaves no stream to probe; still fails on resolution, as before) |

## On the rivestream `not-found` noise — deliberately not "fixed"

The five codes on every successful run are **accurate**: `apex`, `pulse`, `solstice`, `quasar` and `horizon` genuinely return no sources, and `primevids` succeeds. Two reasons to leave them:

1. They cost no latency. `buildRivestreamCycleCandidates` documents that *"every candidate is prefetched in parallel regardless"* — the whole resolve is 1060 ms, the fastest of any provider.
2. They are real signal about which Rivestream mirrors are alive. Suppressing them to quieten the payload would trade a true diagnostic for a cosmetic one.

The reachability guard above is what actually closes the "a real fault could hide in here" concern, since the smoke now fails on evidence rather than on resolution.

## Gates

typecheck 14/14, **5814 pass / 0 fail** (cache-forced), lint identical to the chain baseline (2 + 7 pre-existing, none introduced).

### DIFF
--- a/apps/cli/test/live/allanime-demonslayer.smoke.ts
+++ b/apps/cli/test/live/allanime-demonslayer.smoke.ts
@@ -1,4 +1,5 @@
 import { searchTitles } from "@/services/search/SearchRoutingService";
+import { isStreamReachableForResolve, probeStreamReachability } from "@kunai/providers";
 
 import {
   buildProviderSmokePayload,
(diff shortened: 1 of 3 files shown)


KitsuneKode:
@coderabbitai full review

PR #227, opened 2026-08-25
feat(setup): seven framed screens, honest dependency reporting, real account linking

KitsuneKode:
> Stacked on #224, which is stacked on #223. **Merge order: #223 → #224 → #225.**
> This branch contains both of their commits.

## What

Setup was the only surface in Kunai without the shell frame — no header, no
surface label, no key hints. `.reference/design/cli/kunai-sakura-systems.html`
specified onboarding inside that frame from the start; the implementation
ignored it. Seven screens now, each of which asks something or pays something
off. Three of the old seven asked nothing.

| # | Screen | Decides |
| --- | --- | --- |
| 1 | dependencies | nothing — `[d]` shows the install command for *this* platform, `[r]` re-probes |
| 2 | mode | `defaultMode` |
| 3 | language | audio + subtitles, written to all three lanes |
| 4 | playback | `autoNext`, `skipIntro`, `skipCredits` |
| 5 | downloads & accounts | downloads, quality, AniList / TMDB / Discord |
| 6 | usage ping | consent (unchanged from #224) |
| 7 | done | what was chosen, what is still outstanding |

## Two bugs found by running it, not by reading it

**The footer was invisible.** `SetupFrame` sized itself to `stdout.rows` — but
setup mounts *inside* the app shell's box (`RetainedRootContentLayer` gives it
`flexGrow: 1` and nothing else), and the shell header, status row, and footer
have already taken their share. The frame was taller than its container, so the
footer fell off the bottom of the screen and took every key hint with it. It now
grows into whatever the parent gives. `.docs/design-system.md` already stated
the rule: *the width budget is owned by the container, not the terminal.*

**The account toggles were cosmetic.** Screen 5 wrote `sync.<tracker>.enabled`
and nothing ever opened a browser — a user who asked for AniList got a config
flag and no account. That is the silent no-op this plan exists to remove, not
reproduce. `connectNamedTracker` is now exported, and setup calls it once config
commits — outside the wizard's lifetime, so a browser that never opens costs an
account link rather than the whole setup. Failures are recorded and reported
with the command to retry.

## Honest dependency reporting

**ffprobe is presented as ffmpeg.** No platform ships a package called
`ffprobe`; it arrives inside ffmpeg, so the old *"Install ffprobe from your
platform media-tools package"* was unactionable on all six. The consequence was
also understated — every yt-dlp format selector Kunai uses is a merge
(`bv*+ba/b`, `bestvideo[height<=N]+bestaudio`, `--merge-output-format mp4`), and
without ffmpeg yt-dlp silently falls back to a single progressive stream. So
quality caps rather than anything failing.

**Install instructions are resolved, not listed.** `infra/os/install-commands.ts`
is the single source, and `remediation` lines are *generated* from it rather
than written beside it — a command can no longer be right in one surface and
stale in another. Linux picks from the package manager actually present rather
than what the distro claims to be.

**The shell says what is missing, once.** Previously that was a single
`console.error` the Ink shell painted over milliseconds later.
`CapabilityIssueStrip` derives from the live snapshot at render — never stored,
so re-probing clears it — and is context-gated: no ffmpeg warning when downloads
are off, no curl-impersonate warning outside anime mode, and nothing raised that
has no command to fix it.

## Docs are amended here, not after

The analytics contract, `AGENTS.md`, and the user-facing privacy page are in
this commit because **#224's behaviour already depends on them**. Without this,
the shipped contract says *"setup defaults to off"* while the code pre-selects
on — exactly the doc rot CLAUDE.md says to fix in the same change set.

The doc-drift test that guarded *"off by default"* now guards the narrower and
stronger property: consent is a keystroke, and no skip, accept-all, or
non-interactive path may grant it.

## Verification

| Gate | Result |
| --- | --- |
| `bun run typecheck --force` | 14/14 |
| `bun run lint --force` | 0 warnings, 0 errors |
| `bun run test --force` | 23/23 tasks, 0 fail |
| `bun run build` | 10/10 |
| `bun run verify:doc-paths` | clean |

All with `--force`: a cached replay is not evidence.

**21 committed frame captures** at 72 / 100 / 140 columns, now inside the
per-line width gate. That gate earned its place immediately — the dependency
rows overflowed at 72 columns, and Ink does not truncate an overflowing row, it
*wraps* it, which broke column alignment and inserted blank lines between rows.

**A 25-case platform matrix** asserts every dependency resolves an actionable
instruction on darwin, win32, and Linux with pacman / apt / dnf / zypper /
nothing-recognised. Every prior macOS and Windows CI failure in this stack was a
test pinning one OS's incidental behaviour, so this asserts the property
instead. It also caught that `CURL_INSTALL` had no fallback and resolved to
`null` on a minimal container.

Real binary, isolated profile, no TTY:

```
kunai --setup </dev/null   → exit 1, names the reason, no config written
kunai -S "Dune" </dev/null → exit 1, did not hang, 0 wizard frames
```

## Not done — please read before merging

- **The interactive flow has not been walked through a real PTY.** Screens are
  proven via the render harness and the built binary's guards, not a human
  session. `esc` in particular cannot be exercised: the capture harness does not
  deliver a lone escape byte in a form Ink's `useInput` resolves.
- **Per-lane language configuration is not offered.** One audio and one subtitle
  choice is written to anime, series, and movies alike. Screen 3 does not yet
  *show* that it wrote all three, which reads as though only one lane was
  configured.
- **The brand treatment is thin.** The dot-matrix Kunai mark and the Sakura
  petal motif are not on the welcome screen; the palette is used correctly but
  sparsely. This is the next pass, not a defect in what is here.


KitsuneKode:
Superseded by #235, which merged as `1ee8d09e`.

#235 is a strict superset: it carries this branch's `feat(setup): seven framed screens, honest dependency reporting, real account linking` commit rebased onto `main`, then adds the six audited defect fixes (#228–#233), the analytics skip-path fix, and the pre-setup config restore point. Nothing from this PR is lost — the seven framed screens ship in 0.3.0 through #235.

PR #253, opened 2026-08-26
feat(docs): per-share social cards and server-rendered analytics charts

KitsuneKode:
**Stack 2 of 4** — base `stack/1-docs-context`. Docs site only.

Share links now unfurl with the actual title (verified by rendering the PNG from a built server). The charts are server-rendered SVG with a table twin for every value — `/analytics` is static ISR with no `"use client"`, so a client chart runtime would ship ~100KB and render empty HTML until hydration.

Version is **ordinal, not categorical**: bands take one hue with monotone lightness, both ramps machine-checked. Five bands is a measured ceiling — a sixth fails the adjacent-lightness check.

Live data today is one install with everything suppressed, so the fully-suppressed case is the designed case.

Gates standing alone: typecheck 14/14, lint 12/12, 109 docs tests, doc gates clean.

### DIFF
--- a/apps/docs/app/analytics/page.tsx
+++ b/apps/docs/app/analytics/page.tsx
@@ -1,5 +1,7 @@
+import { TrendSection } from "@/components/analytics/trend-section";
 import { UsagePanel } from "@/components/analytics/usage-panel";
 import { fetchDocsAnalyticsMetrics } from "@/lib/analytics-metrics";
+import { fetchDocsAnalyticsSeries } from "@/lib/analytics-series";
 import { buildPageMetadata } from "@/lib/page-metadata";
 import type { Metadata } from "next";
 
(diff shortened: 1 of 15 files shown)


KitsuneKode:
Split into two independent PRs rather than reopened, because the two halves have no shared files and very different value:

- **#262 — share cards.** Fixes a live defect: every `/w/<code>` link unfurls with the generic card today. 8 files.
- **#263 — analytics charts.** Correct and verified, but renders "below the reporting floor" until a version has 5+ installs. 7 files.

Both based on `main`, so either can merge first.

This PR closed automatically when `stack/3-analytics-series` was deleted on merging #261 — a PR whose base branch disappears gets closed. That was my error: I used `--delete-branch` on a stacked chain, where each PR's base *is* another PR's branch. No content was lost.

PR #254, opened 2026-08-26
feat(analytics): publish the day-by-day series that was already being stored

KitsuneKode:
**Stack 3 of 4** — base `stack/2-docs-site`. `apps/analytics-ingest` only.

`daily_rollup` is never pruned, so the history was already stored while `/metrics/daily.json` served one row of it. New `/metrics/series.json`. **No new collection, no payload change, no consent change.**

Suppression runs across the whole window, not per day — a bucket blinking in and out at the floor is itself a signal about a small population. Absence before launch does not disqualify a version, or every new release would be hidden.

20 tests cover the privacy rules as invariants; 5 run the real path over a memory store.

Gates standing alone: typecheck 14/14, lint 12/12, doc gates clean.

### DIFF
--- a/apps/analytics-ingest/api/metrics/series.ts
+++ b/apps/analytics-ingest/api/metrics/series.ts
@@ -1,0 +1,64 @@
+import type { IncomingMessage, ServerResponse } from "node:http";
+
+import { PUBLIC_METRICS_CACHE_CONTROL, snapshotDayKey } from "../../src/public-metrics.js";
+import { buildPublicSeries, clampSeriesDays, seriesStartDay } from "../../src/public-series.js";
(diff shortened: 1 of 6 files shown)


KitsuneKode:
Closing in favour of #260 — same branch, same content, correct position in the stack.

GitHub auto-marked this "merged" when the stack was reordered: its head (`stack/3-analytics-series`) became an ancestor of its then-base (`stack/2-docs-site`), which GitHub reads as merged. **Nothing reached `main`** — `apps/analytics-ingest/api/metrics/series.ts` is absent from `main` and present on the branch.

The reorder was the point: the endpoint has to deploy and be verified **before** the charts that consume it are worth merging, and this PR previously sat above them.

PR #255, opened 2026-08-26
refactor: clear anti-slop chained assertions from src and make the rule blocking [hold until after 0.3.0]

KitsuneKode:
Rebased onto `main` (was 47 commits behind, branched before #261/#285/#311). Base is `main`; the old "Stack 4 of 4 / base `stack/3-analytics-series`" line is gone — that branch merged as #261 and no longer exists.

All 15 src findings of `no-chained-type-assertions` cleared; the rule now blocks in `.oxlintrc.json` with test paths exempted (~300 baseline there).

Most were not load-bearing — both subtitle casts were unnecessary, including a `selectAutomaticSubtitle(mergedSubtitleList as never, ...)`. `SubtitleEntry` already contains every field of `SubtitleTrack`, so removing them typechecks unchanged. Where the boundary is genuinely untyped (videasy WASM, `Bun.generateHeapSnapshot()`, npm pack metadata) the assertion is now *earned* by a check rather than moved.

Verified the gate bites: reintroducing a chained assertion fails oxlint, removing it passes.

Gates standing alone: typecheck 14/14, lint 12/12, doc gates clean. Full suite 23/23 on the combined tip.

---
**Follow-up commit — `fix(lint): keep the wider test glob to the anti-slop rules only`**

The new rules were added to the override block that already turns off `typescript/no-explicit-any`, `typescript/no-non-null-assertion` and `unicorn/consistent-function-scoping`. Widening that block's glob to `**/test/**` + `**/__mocks__/**` therefore also stopped warning about `any` and non-null assertions in **48** non-test-file `.ts` files under `test/` (helpers, harness, templates, preload). They are warnings, so CI stayed green either way — which is why it would never have been noticed. The wider glob is right for the anti-slop rules; it just should not have dragged the other three along. Now its own override block.

**Gates on the rebased tip** — `lint` 12/12, `typecheck` 14/14, `fmt:check` 26/26, all `0 cached`. Full suite: 5350 pass / 1 fail on one run (`DownloadService youtube argv contract > abort calls runYtDlpProcess cancel handle`), then 3 clean full runs and 3 clean targeted runs — intermittent under the parallel suite, same class as the `buildDiagnosticsInsight` flake, not from this change.

https://claude.ai/code/session_01EfworT7Px6Ky5QkuikrWqG

### DIFF
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -7,8 +7,12 @@
     "es2026": true
   },
   "plugins": ["eslint", "typescript", "import", "unicorn", "promise", "node", "react", "jsx-a11y"],
-
-  "jsPlugins": [{ "name": "anti-slop", "specifier": "./tools/oxlint/anti-slop/index.ts" }],
+  "jsPlugins": [
+    {
+      "name": "anti-slop",
+      "specifier": "./tools/oxlint/anti-slop/index.ts"
+    }
+  ],
   "categories": {
     "correctness": "error",
     "suspicious": "warn",
@@ -19,25 +23,18 @@
     "anti-slop/no-object-parameters": "error",
     "anti-slop/no-reflect-apply": "error",
     "anti-slop/no-reflect-get": "error",
-
     "no-debugger": "error",
     "no-console": "off",
-
     "curly": "off",
     "eqeqeq": "warn",
-
     "no-await-in-loop": "off",
     "no-underscore-dangle": "off",
-
     "eslint/no-unused-vars": "off",
     "typescript/no-unused-vars": "warn",
-
     "typescript/no-explicit-any": "warn",
     "typescript/no-non-null-assertion": "warn",
-
     "import/no-cycle": "warn",
     "promise/no-return-wrap": "warn",
-
     "react/react-in-jsx-scope": "off",
     "unicorn/prefer-node-protocol": "warn",
     "unicorn/no-array-for-each": "off",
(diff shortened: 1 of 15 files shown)


KitsuneKode:
## Hold until after 0.3.0 ships

This is the only PR in the stack that lands in the **CLI binary** — the other three touch `apps/docs` and `apps/analytics-ingest`, which deploy separately and cannot reach an installed `kunai`.

0.3.0 has not shipped yet (latest tag is v0.2.5), so the reasoning is timing, not correctness:

- 10 CLI runtime files change, two of them on the playback path (`PlaybackPhase.ts`, `subtitle-selection.ts`).
- There is **no user-facing benefit** in this PR. It is a type-hygiene refactor plus a lint ratchet.
- If anything odd turns up in 0.3.0, the diff between what was tested and what shipped should be as small as possible. A zero-benefit refactor is exactly what to land *after* a release, not days before one.

It also adds a **new blocking CI gate**. I checked whether that would red anything in flight — #248, #250, #251, and #256 each add **0** new `as unknown as` in `src`, and the uncommitted presence/videasy work on the main checkout is clean under the stricter config too. So the risk is low, but the week of a release is still the wrong week to introduce a gate.

### What the playback changes actually are

Both are cast deletions with no logic moved. `SubtitleEntry` already contains every field of `SubtitleTrack` (the extras are optional), so `SubtitleTrack[]` was always assignable to `readonly SubtitleEntry[]` — the casts were silencing an error that did not exist. One of them was `selectAutomaticSubtitle(mergedSubtitleList as never, ...)`; `as never` is the strongest escape hatch TypeScript has and it was hiding nothing. Casts are erased at compile time, so the emitted JavaScript is identical apart from one removed local.

284 playback and subtitle tests pass, full suite 23/23.

### What actually deserves review here

Not those two lines — the places where I changed the *shape* of the code rather than deleting a cast:

- `packages/providers/src/videasy/direct.ts` — WASM exports are now checked function by function, so a module missing one fails at load with its name instead of as `undefined is not a function` mid-decode.
- `apps/cli/src/infra/clipboard.ts` — Bun's `Subprocess` is now mapped field by field onto the narrower port instead of asserted.
- `apps/cli/src/infra/diagnostics/heap-profiler.ts` — `Bun.generateHeapSnapshot()` is shape-checked once, so a Bun release that changes the node table fails loudly rather than producing an empty histogram.

### Suggested order

Merge #252 → #253 → #254, ship 0.3.0 via #248, then merge this so the ratchet holds the line on everything after.

PR #286, opened 2026-08-31
fix: close silent-failure, cache-poisoning, and episode-numbering gaps

KitsuneKode:
## What changed

Twelve verified bugs across silent failures, a poisoned discovery cache, episode-numbering mismatches, and transport/path handling — each read against the tree before being actioned.

### Silent failures and lost writes

- **`ConfigServiceImpl`** — a debounce that had already fired cleared `savePending` while its store write was still running, so `flushPending()` on the shutdown path (`main.ts:158`) returned early and `process.exit()` could truncate `config.json`. The in-flight write is now tracked and awaited. The existing "store rejection rejects both `save()` and `flushPending()`" test still holds — the `.catch(() => {})` there is unhandled-rejection suppression on a duplicate handle, not error swallowing.
- **`FileStorage`** — `chmod` sat outside the `try`, so a file deleted between the `exists()` check and the chmod rejected with `ENOENT` instead of returning `null`. The corrupt-JSON backup also wrote `""` when the text read failed, destroying the bytes the backup exists to preserve.

### Cache poisoning

`CatalogDiscoveryService` cached whatever a loader resolved for the full 30-minute TTL, and every loader folded failures — including the `AbortError` from navigating away mid-fetch — into `[]`. One offline blip served an empty trending tray until the TTL expired, recoverable only via `clearTrendingCache()`.

Loaders now reject with `DiscoveryUnavailableError`, only a non-empty success is cached, and all seven call sites degrade to an empty tray for display. Failure is visible to the cache layer; the UI still degrades quietly.

### Episode numbering — one rule, four sites

`(entry.season ?? 1) === job.season` normalized only the history side, so absolute-numbered anime compared `1 === undefined` and matched nothing:

- **Retention** — "keep the last 3 watched" retained **zero** for absolute-numbered anime, and cleanup then deleted them.
- **Queue restore** — `sameEpisodeIdentity` returned early when *either* side had an absolute episode, so `13 === undefined` meant an absolute row never matched its own season-relative entry and the resume head was not promoted after a crash.
- Same bug also found in `shouldAutoCleanupOfflineJob` and `isProtectedEpisode`.

All four now share `sameEpisodeNumbering`. The tell was in the existing code: the sort comparator eleven lines below the broken filter *did* normalize the job side.

### Transport and paths

- **Analytics** accepted any scheme from `KUNAI_ANALYTICS_URL` / `analyticsEndpoint`, so `http://` sent `sha256(installId)` in cleartext. Now https-only with a loopback exception, mirroring `normalizeRelayBaseUrl`. A rejected override **stops sending** rather than falling back to the built-in default — rerouting someone's self-hosted telemetry to Kunai's endpoint would be its own consent breach. Contract and user docs updated in the same change set.
- **Relay** used `new URL("/rpc/x", base)`, whose leading slash discards the base's path, so a relay behind a sub-path 404ed into permanent direct fallback — while `normalizeRelayBaseUrl` deliberately *preserves* that path.
- **Download paths** derived a budget in UTF-16 code units and spent it as UTF-8 bytes. Fixing the units exposed that the estimate itself under-corrects (it shares the overrun with a fixed `Season 01` folder that cannot shrink), so the Windows path is now measured and tightened rather than estimated.
- **`merge-timing`** chose on array length, so a degenerate IntroDB `{startMs: 0, endMs: 0}` discarded a valid AniSkip window and the skip prompt never appeared. It now selects on whether a segment can actually drive a skip, matching `normalizeEndSeconds` in `playback-skip.ts`.

### Test isolation (hazard #1)

`FileStorage` built its path map as a **module-level constant**, so `getKunaiPaths()` ran while the module was being imported. Any suite that imported this file — however indirectly — before pointing `HOME`/`XDG_CONFIG_HOME`/`APPDATA` at a sandbox froze the developer's real `config.json` path in, and a later `new FileStorage()` with no explicit paths then wrote the **live profile**.

The map is now built on first use. Explicit paths still win, so existing tests are unaffected; `exists()` keeps answering `false` for an unknown key while read/write/delete keep throwing.

The regression test asserts the condition that used to lose: the module is already imported by the time it runs, and the storage root set afterwards must still decide where the write lands. Verified the write lands in the sandbox and that `~/.config/kunai/config.json` was not touched.

### Tests

25 new assertions; unit suite **5233 → 5258**. Two gaps closed directly rather than through callers: `sameEpisodeNumbering` (now four callers, previously no test of its own) and `offline-sync-policy.ts` (no test file at all, and it decides whether a downloaded file is deleted).

## Seams walked (`AGENTS.md`)

- **Declaration → reader** — every loader that changed from returning `[]` to rejecting has its consumers updated; all seven discovery call sites verified.
- **Both lanes** — the numbering rule is exactly the anime-vs-TMDB split; absolute and season-relative are covered on both sides and a symmetry test pins it.
- **Every platform** — the path fix is Windows-specific and asserted with CJK and astral cases; POSIX keeps the byte cap.
- **Docs** — `.docs/analytics-privacy-contract.md` and `docs/users/reliability-and-privacy.mdx` updated with the analytics change. `verify:doc-paths` passes.
- **Every provider** — no provider-shaped behavior changed; the relay fix is transport-level and applies to all relay-safe providers.

## Not in scope

Deliberately excluded: `PlaybackPhase` / `shell-workflows` decomposition, the 16 baselined layer inversions, and the legacy `tmdb.ts`/`search.ts` cycle — plans **010/011/012/014/015** own those and are dependency-ordered.

Nine reported findings were checked and refuted, producing no change. The notable ones: download naming *is* asserted (`download-service.test.ts:65` pins `Frieren - E03.mp4`), `vidrock` was already migrated off `crypto-js` to `node:crypto`, the reserved-name guard runs on the bare title before any extension exists, and the TMDB season fan-out is bounded by a filter that admits only seasons with no air date.

## Checklist

- [x] Changeset added — patch
- [x] `bun run guard` passes — `@kitsunekode/kunai@0.3.0` in sync
- [x] `bun run fmt && bun run lint && bun run test && bun run typecheck` pass locally (forced, not cache replays: typecheck 14/14 and test 23/23 with `0 cached`)
- [x] `bun run build` passes — host binary written and `--help` smoked
- [x] Live provider / Discord smokes skipped intentionally — no provider or presence behavior changed

## Note for the reviewer

One pre-existing flake seen: `buildDiagnosticsInsight > healthy session` failed once in six `--parallel` runs, then passed 5/5. Clean `main` is green, and it passes in isolation. Unrelated to this branch but worth a look.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved shutdown reliability by ensuring pending settings are saved before exit.
  - Discovery and trending sections remain usable when catalog services fail.
  - Fixed episode resume, offline cleanup, and absolute-numbered anime matching.
  - Preserved intro-skip timing when preferred metadata is unusable.
  - Improved Windows filename handling and relay URLs hosted under sub-paths.
  - Improved handling of missing, unreadable, and sandboxed configuration files.
  - Storage locations now respect supported home-directory environment settings.

- **Privacy & Documentation**
  - Analytics overrides require HTTPS, except for local development addresses; unsafe endpoints send nothing.
  - Updated privacy, analytics, and contributor guidance documentation.

### DIFF
--- a/.changeset/tidy-pandas-shake.md
+++ b/.changeset/tidy-pandas-shake.md
@@ -1,0 +1,14 @@
+---
+"@kitsunekode/kunai": patch
+---
+
+Fix a set of silent failures found while auditing the tree.
+
+- Quitting immediately after changing a setting could lose the change: a debounced save that had already started was not awaited by the shutdown flush.
+- A single offline blip or navigating away mid-load could leave trending empty for 30 minutes, because a failed discovery fetch was cached as "no results".
+- "Keep the last N watched" retained nothing for absolute-numbered anime, and the same numbering mismatch stopped watched anime downloads from ever being reclaimed and stopped the resume position being restored after a crash.
+- Analytics now refuses a non-https endpoint override instead of sending in cleartext. A rejected override stops sending entirely rather than falling back to the built-in URL.
+- A self-hosted relay behind a sub-path (`https://host/prod`) is now called correctly instead of 404ing and silently falling back to direct.
+- Long CJK and emoji titles could still produce a Windows path over `MAX_PATH`; the path is now measured and tightened rather than estimated.
+- An intro-skip window from AniSkip is no longer discarded when IntroDB answers with an untimed segment.
+- A config file that is unreadable, or that vanishes mid-read, no longer throws or overwrites its own corrupt-backup with an empty file.
--- a/.changeset/tidy-pandas-shake.md
+++ b/.changeset/tidy-pandas-shake.md
@@ -12,3 +12,4 @@ Fix a set of silent failures found while auditing the tree.
 - Long CJK and emoji titles could still produce a Windows path over `MAX_PATH`; the path is now measured and tightened rather than estimated.
 - An intro-skip window from AniSkip is no longer discarded when IntroDB answers with an untimed segment.
 - A config file that is unreadable, or that vanishes mid-read, no longer throws or overwrites its own corrupt-backup with an empty file.
+- The config path is now resolved when it is first needed rather than when the module loads, so a sandboxed run can never fall back to the real profile.
--- a/.docs/agents/audit-findings-bar.md
+++ b/.docs/agents/audit-findings-bar.md
@@ -1,0 +1,99 @@
+---
+status: current
+lastReviewed: "2026-08-31"
+---
+
+# The bar for an audit finding
+
+> Agent-facing (L3). Never linked from published docs.
+
+Four audit passes over this repo produced ~38 distinct claims. Roughly a third
+dissolved when someone opened the twenty lines around the cited one. The line
+numbers were nearly always right; the inference from them was not.
+
+This file is the bar. Apply it before filing, and when reviewing findings
+someone else filed.
+
+## The recurring failure
+
+**A true observation about a line, plus a false inference about what that line
+does.** Every refuted finding had this shape:
+
+| Filed as                                                               | What one more hop showed                                                                                   |
+| ---------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------- |
+| "Empty catch swallows read errors, watchdog stops protecting"          | The catch is inside a template literal that becomes Worker source, and the fallback is a valid RSS reading |
+| "N+1 fan-out: dozens of concurrent TMDB calls for long-running series" | The list is filtered to seasons with _no air date_; One Piece exits before the `Promise.all`               |
+| "Naming regressions would pass — `Frieren - E03.mp4` is untested"      | That exact string is asserted in `download-service.test.ts:65`                                             |
+| "crypto-js on the hot decrypt path in `vidrock/direct.ts:2050`"        | That file is 124 lines and already uses `node:crypto`; a test pins the migration                           |
+| "`NUL.mp4` bypasses the reserved-name guard"                           | Sanitisation runs on the bare title, before an extension exists                                            |
+| "Config save failures are swallowed"                                   | A test named "store rejection rejects both save() and flushPending()"                                      |
+
+## Six rules
+
+1. **Read the callee, not the call site.** If a finding depends on what a
+   predicate admits, a helper returns, or a constant holds — open it. Two of the
+   refutations above died on the function one hop away.
+
+2. **Grep the test file before claiming something is untested.** Cheapest check
+   available, and three findings would not have been filed.
+
+3. **An empty `catch` is a finding only if it is undefended.** Quote the comment
+   above it, or state there isn't one. This codebase comments its deliberate
+   silences — `memory-watchdog.ts:72`, `atomic-write.ts:135`,
+   `ConfigServiceImpl.ts:758` all carry their rationale.
+
+4. **State the disproof you attempted.** Not a confidence label — _what did you
+   check that would have made this wrong?_ Across four passes, stated confidence
+   had no relationship to whether a finding survived: a HIGH-confidence item
+   cited line 2050 of a 124-line file. Treat the label as noise; read the
+   evidence.
+
+5. **Check `.plans/roadmap.md` before filing anything architectural.** Four
+   findings in one pass restated plans 010/011/012/014/015, which are already
+   written, prioritised, and dependency-ordered with a stated reason for being
+   BLOCKED.
+
+6. **Reproduce; do not reason.** Especially for anything platform-shaped. See
+   below — this one is written in the author's own blood.
+
+## Reproduce, do not reason
+
+The `FileStorage` lazy-path test in this same change set failed CI **three
+times**, each for a different host-dependent assertion the author reasoned their
+way into:
+
+- hard-coded that the sandboxed config lands under the temp dir (macOS puts it
+  under `Library/Application Support`, Linux under XDG, Windows under APPDATA);
+- compared path _spelling_ — macOS reports `/var/...` for a directory resolving
+  to `/private/var/...`, Windows reports the 8.3 form (`RUNNER~1`);
+- read a file with `.catch(() => null)` and asserted `not.toContain` on it,
+  which throws on a null receiver — passing only on a machine that happens to
+  have a real `config.json`.
+
+Each was "obviously fine" by inspection. The fix each time was to simulate the
+condition instead: an empty `HOME` with XDG and APPDATA unset reproduced the
+third failure locally in one command.
+
+Corollary: **a cross-platform assertion you have not run on that platform is a
+guess.** Prefer assertions that hold everywhere — compare against the same
+resolver production calls, or assert an inequality, rather than a path shape.
+
+## Parallel agents
+
+Two sweeps that start from the same grep produce the same blind spot twice.
+Agreement between them is duplication, not corroboration. Deduplicate before
+reporting, and never raise confidence because two agents said it.
+
+## What a filed finding must carry
+
+- The failing input or state, and the wrong output — concretely.
+- `file:line` for the defect **and** for the code that makes it reachable.
+- What you checked that would have falsified it.
+- Whether a test already covers it, and if so why it does not catch this.
+- Whether `.plans/` already owns it.
+
+## Related
+
+- [issue-tracker.md](./issue-tracker.md) — where findings become issues
+- [triage-labels.md](./triage-labels.md) — labels for the result
+- [../../AGENTS.md](../../AGENTS.md) — the four hazards and the "hit every seam" list
--- a/.docs/analytics-privacy-contract.md
+++ b/.docs/analytics-privacy-contract.md
@@ -44,6 +44,16 @@ Read this before touching `apps/cli/src/services/analytics/`,
   `analyticsEndpoint` (the shipped default) means “use the built-in URL”, not
   “disable sending”. Disable in Settings, or with `DO_NOT_TRACK` / `CI` / a
   non-TTY session.
+- **An override must be https.** `http://localhost`, `127.0.0.1`, and `[::1]`
+  are the only cleartext exceptions, so `apps/analytics-ingest` can be developed
+  against locally. Anything else — an `http://` host, a non-URL — is refused and
+  **stops sending entirely**; it is never redirected to the built-in default,
+  because someone who pointed their installs at their own ingest did not consent
+  to sending here instead. The wire carries `sha256(installId)` with version,
+  OS, and architecture: over http that is readable on the path and one install
+  is followable day to day, which is exactly what the digest exists to prevent.
+  Same rule and same loopback exception as `normalizeRelayBaseUrl` in
+  `packages/relay`.
 - The default is a domain Kunai controls, never a hosting provider's own URL.
   This string is baked into immutable npm tarballs and compiled binaries, so it
   must outlive whatever serves it today: DNS can be re-pointed, a published
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -204,6 +204,10 @@ legacy implementation details.
 Issues are GitHub Issues on `KitsuneKode/kunai` — workflow in
 [.docs/agents/issue-tracker.md](.docs/agents/issue-tracker.md), labels in
 [.docs/agents/triage-labels.md](.docs/agents/triage-labels.md). Domain language
-is [.docs/agents/domain.md](.docs/agents/domain.md). Kunai is single-context:
+is [.docs/agents/domain.md](.docs/agents/domain.md). Before filing an audit
+finding — or acting on one someone else filed — read
+[.docs/agents/audit-findings-bar.md](.docs/agents/audit-findings-bar.md): about a
+third of the findings this repo has received were true about a line and wrong
+about what it does. Kunai is single-context:
 one system-wide ADR set in `.docs/adr/`, no per-package context files, and new
 ADRs take the next sequential number.
--- a/apps/cli/src/app/discover/discover-sections.ts
+++ b/apps/cli/src/app/discover/discover-sections.ts
@@ -7,6 +7,7 @@
 
 import { loadDiscoveryList } from "@/app/discover/discovery-lists";
 import type { Container } from "@/container";
+import type { SearchResult } from "@/domain/types";
 import {
   historyContentType,
   readLatestHistoryByTitle,
@@ -65,11 +66,13 @@ export async function buildDiscoverSections(
           .getForTitle(recentHistoryId, historyContentType(mostRecentCompleted[1]))
           .then((s) => ({ ...s, label: `Because you watched ${mostRecentCompleted[1].title}` }))
       : null,
-    loadDiscoveryList(mode).then((items) => ({
-      label: mode === "anime" ? "Anime trending this week" : "Trending this week",
-      reason: "trending" as const,
-      items,
-    })),
+    loadDiscoveryList(mode)
+      .catch((): readonly SearchResult[] => [])
+      .then((items) => ({
+        label: mode === "anime" ? "Anime trending this week" : "Trending this week",
+        reason: "trending" as const,
+        items,
+      })),
     !options.light && mode !== "anime" && completedHistorySeeds.length > 0
       ? container.recommendationService
           .getPersonalizedByHistory(completedHistorySeeds)
--- a/apps/cli/src/app/search/SearchPhase.ts
+++ b/apps/cli/src/app/search/SearchPhase.ts
@@ -819,7 +819,13 @@ export class SearchPhase implements Phase<SearchPhaseInput | void, TitleInfo> {
             // the empty-state copy below rather than tearing down the phase.
             const results = await observeOnline(container, "search-error", () =>
               loadDiscoveryList(mode, context.signal),
-            ).catch((): SearchResult[] => []);
+            ).catch((error: unknown): SearchResult[] => {
+              // A cancelled navigation is not an empty tray. Swallowing it here
+              // recorded a "succeeded / healthy" discovery event and pushed state
+              // into a phase the user had already left.
+              if (context.signal.aborted) throw error;
+              return [];
+            });
 
             logger.info("Discovery list loaded", {
               mode,
@@ -1244,7 +1250,12 @@ async function loadSearchRoute(
   const bundle =
     route === "trending"
       ? {
-          results: await loadDiscoveryList(mode, context.signal).catch((): SearchResult[] => []),
+          results: await loadDiscoveryList(mode, context.signal).catch(
+            (error: unknown): SearchResult[] => {
+              if (context.signal.aborted) throw error;
+              return [];
+            },
+          ),
           subtitle:
             mode === "anime"
               ? "AniList trending"
--- a/apps/cli/src/app/search/SearchPhase.ts
+++ b/apps/cli/src/app/search/SearchPhase.ts
@@ -814,9 +814,12 @@ export class SearchPhase implements Phase<SearchPhaseInput | void, TitleInfo> {
             stateManager.dispatch({ type: "SET_SEARCH_QUERY", query: "" });
             stateManager.dispatch({ type: "SET_SEARCH_STATE", state: "loading" });
             const mode = stateManager.getState().mode;
+            // Trending rejects when the upstream is down, which is what lets
+            // `observeOnline` see the failure at all. Browse still degrades to
+            // the empty-state copy below rather than tearing down the phase.
             const results = await observeOnline(container, "search-error", () =>
               loadDiscoveryList(mode, context.signal),
-            );
+            ).catch((): SearchResult[] => []);
 
             logger.info("Discovery list loaded", {
               mode,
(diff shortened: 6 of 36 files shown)


KitsuneKode (commented):
# Review #286 — fix: silent failures, cache poisoning, episode numbering, macOS isolation

**Worktrees:** `.claude/worktrees/review-286` (commit `efad9573`), `origin/main` for diff  
**Gates run (forced, 0 cached):** `typecheck --force` 14/14 pass, `lint` 12/12 pass, `verify:doc-paths` 51 docs pass, `test --force` timed out on turbo parallelism in this env but `providers` shard passes standalone (544 pass); `build --help` smoked by author. The `buildDiagnosticsInsight > healthy session` flake seen on #289/#293 is pre-existing (1/6 parallel runs in audit prompt) — two hits in my two runs are same signature, not this PR.

## Verdict: MERGE AFTER FIXES (2 nits, 0 blockers)

All 8 areas were re-verified against the tree, not the PR description. No crash, no data-loss reversal found.

---

### 1. Discovery loaders rejecting vs `[]` — all 7 call sites do degrade, aborts are correct

**Claim:** "All seven call sites handle rejection" — **TRUE**.

`grep -rn "loadDiscoveryList\|loadTrending(\|loadSurprise(\|loadYoutubeTrending"` across `review-286` gives:

* `services/catalog/CatalogDiscoveryService.ts:70,88` — loaders themselves, now `throw DiscoveryUnavailableError` on transport/non-2xx/bad JSON; `runLoad` only caches `length>0` and propagates rejection.
* `app/discover/discovery-lists.ts:13,20` — thin wrappers.
* `app/discover/discover-sections.ts:69` — `loadDiscoveryList(mode).catch((): SearchResult[]=>[])` → empty tray, correct.
* `app/search/SearchPhase.ts:821` (onLoadDiscovery via `observeOnline`) — `.catch(error=> signal.aborted? throw : [])` → abort re-thrown, otherwise empty. `observeOnline` records failure then re-throws, so the catch sees the same error. **No crash:** the re-thrown abort propagates out of `onLoadDiscovery`; its caller is the browse inline-loop (`SearchPhase.ts:813` is the `onLoadDiscovery` handler invoked by the browse shell). The shell's dispatch path swallows `AbortError` via the surrounding `if (context.signal.aborted) throw error` at `SearchPhase.ts:901` in the sibling `onLoad` — same pattern, and the top-level `SearchPhase` loop has an outer `try/catch` that contains the failure to `search.query.failed`. Manual trace shows an aborted navigation never reaches the `SET_SEARCH_RESULTS` dispatch that would have recorded a healthy discovery event.
* `app/search/SearchPhase.ts:1253` (`loadSearchRoute` trending) — same `signal.aborted? throw : []`.
* `app/discover/random-results.ts:22,23,30,36,50,51` — both `loadDiscoveryList` and `loadSurpriseList` wrapped with `.catch(()=>[])` for `/random` and `/surprise`. `/random`'s `loadDiscoverResults` light path also catches.
* `services/youtube/YoutubeRecommendationService.ts:221` — `loadYoutubeTrending` now bare `await invidiousGetTrending` (no catch). Its *other* caller at `~211` (`loadYoutubeRecommendations` trending fallback) still has `.catch(()=>[] as ProviderSearchResult[])` — confirmed at `review-286:YoutubeRecommendationService.ts:211-216`.

**Harder question:** Does the user ever see ERROR where they previously saw empty tray? No — every UI entry that previously got `[]` still gets `[]` via the catches. The only new error path is `CatalogDiscoveryService` cache layer seeing the rejection, which is intentional so the next call retries instead of serving the poisoned TTL. No unhandled rejection crashes browse.

**Test that would have falsified:** remove the `.catch` in `discover-sections.ts:69` and assert the tray renders empty — currently passes only with the catch.

---

### 2. `ConfigServiceImpl` in-flight handle

**Fixed correctly** at `ConfigServiceImpl.ts:800` (`this.saveInFlight = pending` assigned *before* `store.save` starts). The CodeRabbit-found bug (assigned after) would have left a rejected pending stranded.

Checked orderings:
* **Two overlapping `save()`** — coalesced into one `savePending` promise via debounce; `persistPendingSave` clears `savePending`/`Resolve/Reject` then sets `saveInFlight=pending` before the async write, so the shared promise settles for both callers.
* **`flushPending()` racing a debounce that fired mid-await** — `flushPending` at `773` first checks `savePending` (still present if timer hasn't fired) else awaits `saveInFlight`. If the timer fired between the check and the await, `saveInFlight` is already set, so it awaits the in-flight write instead of returning early — the truncation bug.
* **`reset()` while write in flight** — `reset()` at `~815` does `await store.save` directly, bypassing the debounce. Concurrent `saveInFlight` and `reset` could interleave two `store.save` calls ordering by who finishes last. Not a hang: `saveInFlight` is cleared in `finally` only if still `===pending`, so a racing `reset` doesn't strand it. No dead hang where `saveInFlight` holds a promise nothing settles.

**Falsification:** stub `store.save` to never resolve, call `save()` then `flushPending()` — without the fix `flushPending` returned immediately; with fix it hangs awaiting `saveInFlight` as intended (verified by reading the `finally` guard).

---

### 3. Analytics https-only + `redirect:"error"`

`usage-analytics-service.ts:121-126` + `resolveAnalyticsEndpoint:acceptOverride` — rejected override returns `""`, and `maybePing:164` returns early on `!endpoint` and on `!isTransportSafeAnalyticsEndpoint(endpoint)` second gate.

* **Empty endpoint re-enabling default?** No. `resolveAnalyticsEndpoint` is called at startup to inject `endpoint` into the service; an empty string is injected, not replaced. No fallback path substitutes `DEFAULT_ANALYTICS_ENDPOINT` when `endpoint===""` — the default is only used when *no* override was provided. Verified no `if (!endpoint) endpoint = DEFAULT` downstream.
* **Three architecture tests:** `analytics-disclosure-once`, `analytics-endpoint-pin`, `analytics-payload-drift` still assert what they claim — checked `analytics-privacy-contract.md` now documents the `https-only + "" stops sending` posture, consistent with code.
* **`redirect:"error"` turning 3xx into permanent disable?** Opposite: `send` treats `status>=500` as `retry`, else `permanent`. With `redirect:"error"` a 302 throws, caught as `retry` (line `catch { return "retry" }`), so a legitimate ingest that 302s retries rather than consuming the 24h cadence. Without the guard, https→http redirect would have leaked `installId` digest in cleartext. The choice is correct.

**Nit:** `isTransportSafeAnalyticsEndpoint` allows `http://[::1]` and `http://::1` but `http://[::1]:3000` hostname is `"[::1]"` (brackets kept by `URL`), so `[::1]` set matches. Verified the loopback test in `usage-analytics-service.test.ts` passes.

---

### 4. `download-path-naming.ts` Windows budget loop

* **Empty stem / dropped extension?** No: `clampComponent:131` does `stripTrailingDotsAndSpaces(truncateToLimits(...)) || "file"` so empty becomes `"file"`, extension preserved via `budget` split.
* **32-iteration bound reachable? What at floor 16?** Loop at `~205-212` tightens `maxUtf16` by `ceil(overrun/variableCount)`. Floor is `Math.max(16, ...)`. When `next===limits.maxUtf16` (would not tighten) it breaks and "lets the write surface the real error rather than silently mangling" — comment at `~210`. So deeply nested `baseDir` beyond `WINDOWS_MAX_PATH - RESERVE` stops tightening; no infinite loop, no empty output.
* **`clampComponent` signature `number|ComponentLimits` — stale numeric callers?** Verified all call sites: `resolveDownloadOutputPath:196` uses `limits:ComponentLimits` from `componentBudget`; POSIX `componentBudget` returns `{maxBytes:255, maxUtf16:Infinity}` so numeric call path is only via `clampComponent("Dune.mp4",255)` in tests and one internal `truncateToBytes` compat alias removed. No stale numeric caller now under-constrains Windows because Windows path always goes through `ComponentLimits` + the 32-iteration tightener. The change description "bytes only, unlimited UTF-16" is exactly the POSIX semantics.

**Falsification:** revert `componentBudget` to return `number` and run `Windows path budget units > long CJK path stays inside MAX_PATH` — fails by exceeding 248.

---

### 5. `packages/storage/src/paths.ts` HOME handling

* **`HOME=""` falsy falls through:** `env.HOME ?? env.USERPROFILE ?? homedir()` — `""` is not `null/undefined`, so `??` would keep `""`. But code at `review-286` is `env.HOME ?? env.USERPROFILE ?? homedir()` — empty string is truthy for `??`, so would keep `""`. **However** the *effective* handling is correct because `join("", "Library/...")` would produce a relative path, but the test `paths-home-resolution.test.ts` never covers empty string. I checked actual runtime: `env.HOME` empty string would produce `home=""` → `join("", "Library")` = `"Library/Application Support/kunai"` (relative). The prompt's premise "empty string is falsy so it falls through; confirm" is *false* for `??` — empty string does **not** fall through. **Is this a bug?** In practice `HOME=""` is not set by any OS; an unset `HOME` is `undefined`, not `""`. A shell `export HOME=""` is pathological. The docs' hazard #1 remediation targets sandboxing with `HOME=/tmp/...`, not empty. Risk is LOW — no real user hits this, and the prior `homedir()` fallback already had the same issue if HOME unset. If you want it defensive, use `env.HOME?.trim()? env.HOME : env.USERPROFILE?...`.

* **`HOME` relative path → profile moves to relative dir?** Yes, `join(home, "Library/...")` with relative `home` yields relative `configDir`. A process launched with `HOME=relative` would resolve relative to cwd. This is the platform's own semantics; no guard is expected. Not a defect.

* **Windows service account USERPROFILE differs from real profile?** `USERPROFILE` is the documented per-user profile; service accounts have distinct profiles — honoring `USERPROFILE` is correct. The change description notes `homedir()` reads account record rather than env, which is Bun-specific and broke macOS sandboxing.

* **Long-lived process caching?** `FileStorage` now caches `resolvedPaths` per instance (`this.resolvedPaths ??= this.paths ?? defaultPaths()`) after first `lookupPath:109`. That's intentional per-instance memo, not global. `getKunaiPaths` itself is stateless each call. So a process that mutates env after creating `FileStorage` keeps the first resolution for that instance — expected; re-creating the service picks up new env.

---

### 6. `sameEpisodeNumbering` — one rule, four sites

At `domain/media/episode-numbering.ts:15-26` — "absolute wins only when BOTH sides carry one".

* **Checked old per-site behaviours:** offline keep set `(entry.season??1)===job.season` normalized only history side; queue `sameEpisodeIdentity` early-returned when *either* side had absolute; cleanup and protected-episode each normalized differently. All four now share the rule.
* **Probe absolute 13 vs S02E01:** `sameEpisodeNumbering({absoluteEpisode:13},{season:2,episode:1})` → `false` (test at line `season beyond first is not assumed` pins this). The old per-site code would also have been `false` for different reasons, but the new rule is *conservative* — without season-length metadata, assuming 13 == S02E01 when S1 has 12 episodes would be a false positive (merging unrelated episodes), worse than a false negative (missing a match). The existing tests document this tradeoff explicitly.
* **Title where S1 has 12 episodes:** If a show indeed has 12-ep S1, then absolute 13 *is* S02E01, and the shared rule would return `false` incorrectly. No code in the repo knows per-season episode counts, so any rule without that data must choose. The author's choice is the safer under-merge.

No blocking defect.

---

### 7. `FileStorage.ts` read() → null on permission/IO

Callers: only `ConfigStoreImpl:17` (`this.storage.read<unknown>(STORAGE_KEY)`) and `FileStorage.test` itself. `ConfigStoreImpl.load` does `const raw = await storage.read(...); if (!raw) return DEFAULT_CONFIG` plus migration — returning `null` on unreadable with a `warn` ("Config file could not be read; defaults are in use") is the intended degrade: no throw to distinguish "unreadable" vs "absent" is needed because both result in defaults + warning. No caller previously relied on the throw to branch differently — verified via grep, only `FileStorage` consumers are persistence.

---

### 8. `merge-timing.ts` usable test

`preferUsable` checks `isUsableSegment: endMs>0 finite`. `playback-skip.ts` consumes via `normalizeEndSeconds` (`value>0 finite`) *and* `endSeconds>startSeconds` plus `positionSeconds < end-0.25`. The merge test omits the `start<end` check. A pathological `[{startMs:9999,endMs:1}]` would be deemed usable by merge but dropped by `findPlaybackSegmentAtPosition`. In practice IntroDB never emits inverted segments (it normalizes via `normalizeSegments` in `introdb.ts`), and the degenerate case the bug fixed is `{0,0}` — both agree. The mismatch is narrow and not user-visible. Documented as such.

---

### Cross-note: `clampComponent` call-site search

All `clampComponent` calls are in `download-path-naming.ts:196,215` and tests. No stale numeric call site silently under-constraining Windows.

### Re-verified claims from prompt

* Discovery seven sites — **verified true** (above).
* Overall gates — forced green on typecheck/lint/doc-paths; tests hit known flake not this PR.

**Model:** Muse Spark (Meta) — see `/tmp/opencode/review-bodies/model-info.txt`

KitsuneKode (commented):
## Review (independent of the PR author's self-review)

Cross-PR first: this branch does not touch `packages/storage/src/repositories/history.ts` or `.plans/roadmap.md`. No interaction with #289/#293 or the plan PRs. The 0.3.0 merge-train board (`.plans/2026-08-24-0-3-0-merge-train.md`) does not mention this PR and imposes no merge order.

Gates I ran in `.claude/worktrees/audit-fixes` (`0 cached`): typecheck 14/14, lint 0 warnings, `verify:doc-paths` 51 docs / all resolve. `bun run test --force` failed 2 unit tests; both look pre-existing (the known `buildDiagnosticsInsight` flake, plus an activation-lock reclaim flake that also failed on #293 which does not touch paths). Not treated as this PR.

Mutation check: reverting `sameEpisodeNumbering` to “either side has an absolute → compare only absolute” made `episode-numbering.test.ts` fail at the S01E13 twin case. That test is a real regression test.

---

### Blocking

**1. `HOME=""` does not fall through; it relocates the entire profile to a CWD-relative directory.**

Failing input: `HOME=""` (unset `homeDir`, no `USERPROFILE`).

Wrong output: `getKunaiPaths({ platform: "darwin", env: { HOME: "" } })` sets `home` to `""` because `??` only skips `null`/`undefined`. `posix.join("", "Library", "Application Support", "kunai")` is `"Library/Application Support/kunai"` — a relative path. Linux without XDG is `".config/kunai"`. Every subsequent config/data/cache SQLite file follows it.

Defect: `packages/storage/src/paths.ts` (`const home = options.homeDir ?? env.HOME ?? env.USERPROFILE ?? homedir()`).
Reachable from: every `getKunaiPaths()` call, including `FileStorage.defaultPaths()` on first use and `applyStorageRootEnv`'s production twin.

What I checked that would have falsified it: `"" ?? "fallback"` is `""` (confirmed in Bun). The new tests in `packages/storage/test/paths-home-resolution.test.ts` never pass `HOME: ""`; they only pass non-empty sandboxes. An existing test should have caught this and did not, because none asserts empty-vs-unset.

This is the same class of host-env reasoning this PR already burned on three times. Fix: treat empty `HOME`/`USERPROFILE` as unset (`(env.HOME && env.HOME.length > 0) ? env.HOME : …` or equivalent). Add a case to `paths-home-resolution.test.ts`.

A relative non-empty `HOME` (`HOME=foo`) also moves the profile; that is operator error. Empty is accidental and common in stripped environments.

---

### Non-blocking (checked, not merge-blocking)

**2. `FileStorage.read()` now maps every non-ENOENT IO error to `null`.**

Old: `chmod` sat outside the `try`; EPERM/EACCES threw. New: they return `null` and warn (`FileStorage.ts` catch after chmod/text).

Only production reader: `ConfigStoreImpl.load` (`if (!raw) return {}`), then `ConfigServiceImpl.load` merges `DEFAULT_CONFIG`. A later `save()` / `flushPending()` will write those defaults over the still-present file if the directory is writable.

I did not find a caller that still *needs* the throw to distinguish “unreadable” from “absent” — they all already treated `null` as empty. The overwrite-on-next-save is the residual risk, for a file you could not chmod but might still replace via atomic write. Narrowing the catch to `ENOENT` (the bug you actually had) would close it.

**3. Discovery loaders reject; the seven `loadDiscoveryList` / `loadYoutubeTrending` UI sites degrade to `[]`.**

Verified call sites:

| Site | Catch |
| --- | --- |
| `SearchPhase.ts` `onLoadDiscovery` | `[]`, rethrow if `context.signal.aborted` |
| `SearchPhase.ts` `loadSearchRoute` trending | same |
| `discover-sections.ts:69` | `[]` (no signal) |
| `random-results.ts` (3 `loadDiscoveryList` + surprise) | `[]`, no abort rethrow |
| `post-playback-recommendations.ts:317` | `[]` |
| `discover-results.ts:97` `loadYoutubeTrending` | `[]` |

`loadYoutubeTrending`'s other caller (`YoutubeRecommendationService.ts` ~211, `loadYoutubeRecommendations` trending fallback) still has `.catch(() => [])`.

Abort rethrow does **not** crash browse. `onLoadDiscovery` is awaited inside `browse-shell.tsx` `loadDiscovery`'s try/catch. `loadSearchRoute` abort is caught by `SearchPhase.ts:1205-1206` and returns `{ status: "cancelled" }`. Non-abort failures become an empty tray, not an error surface, except Browse's catch which can show “Trending failed.” if abort fires while the request is still current.

Residual: `/random` and `/surprise` still swallow `AbortError`, so `loadSearchRoute` can dispatch `SET_SEARCH_STATE ready` after the user has left. Same class as the CodeRabbit trending fix, sibling routes only.

**4. `sameEpisodeNumbering` vs absolute 13 / S02E01 / 12-ep season 1.**

I could not find a pair where the *old* per-site behaviour was correct and the new shared rule is wrong. Absolute 13 vs S02E01 is `false` (pinned). Absolute 13 vs S01E13 is `true` (pinned) — wrong for a 12-episode season 1, but the old queue-restore early-return also failed to match the intended twin on 1-season shows. Documented limitation, not a regression.

**5. Other items from the brief, checked and not wrong.**

- `saveInFlight` is assigned before `store.save`. Two overlapping `save()` calls share one pending until persist starts; a `save()` during in-flight creates a new pending. `flushPending` prefers that pending over `saveInFlight`, so shutdown can start a second write — `FileStorage.writeLock` serializes it. `reset()` still does not await `saveInFlight` (pre-existing shape). `saveInFlight` cannot hold a promise that never settles unless `store.save` itself hangs.
- Rejected analytics override → `""`. `maybePing` returns on `if (!endpoint)` (`usage-analytics-service.ts` ~225). Does not fall back to the default. `analytics-endpoint-pin` still asserts the baked-in URL. `redirect: "error"` is caught as `"retry"`, not a permanent disable.
- `clampComponent` callers: only `download-path-naming.ts` itself (passes `ComponentLimits`) and tests (still pass a number). 32-iteration loop is deterministic; floor 16; empty stem becomes `"file"`; extension kept. 32 is not a realistic bound given how fast `overrun/variableCount` drops.
- `merge-timing` `isUsableSegment` matches `normalizeEndSeconds` (`endMs > 0`). `playback-skip.ts` also requires `endSeconds > startSeconds`. `introdb.ts` `normalizeSegments` does not drop `{startMs: 5000, endMs: 5000}`; that would count as usable here and be skipped later. The reported `{0,0}` case is handled.
- `FileStorage.resolvedPaths` is per-instance, first use. Production constructs one at bootstrap; env changes after that are not a real session concern.

Nine earlier refutations: I did not reopen ones this PR already dropped (download naming string, vidrock crypto-js, reserved-name, TMDB fan-out). A-14 / tracker-sync cour-relative absolute: `.docs/tracker-sync.md:79-81` still says an absolute-only number resolves to nothing rather than being guessed. I am not overturning that.

---

Change it: treat empty `HOME`/`USERPROFILE` as unset, and add a test that would have failed today (`HOME: ""` on darwin and linux-without-XDG must equal `homedir()`-based layout, not a relative path).

KitsuneKode:
Both nits fixed in `68ecaf34`, and the first was a correction to my own reasoning rather than to the code as reported.

**`HOME=""`** — you were right and my premise in the review prompt was wrong. I wrote "empty string is falsy so it falls through; confirm", and `??` does not do that: `""` is not nullish, so it survived the chain and became the storage root, making every joined path relative to the working directory. `firstNonEmpty` now trims and skips, so empty or whitespace-only means unset. Two tests added covering `HOME=""` falling through to `USERPROFILE`, and whitespace-only falling through to `homedir()`.

Low practical risk, as you said — no OS ships an empty `HOME`. But the entire purpose of that change was to make sandboxing reliable, and a sandbox that sets the variable badly should fall through rather than land somewhere relative.

**`isUsableSegment`** — also correct. It checked only `endMs > 0` while `findPlaybackSegmentAtPosition` additionally requires the window to move forwards, so `{startMs: 9999, endMs: 1}` counted as timing during the merge and was then dropped by the player. That is precisely the "present but unusable" shape the degenerate `{0, 0}` case caused, which is the failure this rule exists to prevent. My commit message claimed the check "mirrors `normalizeEndSeconds`"; it mirrored half of it. Now requires `endMs > startMs`, with a test using your exact inverted example.

Gates re-run forced, `0 cached`: typecheck 14/14, lint clean, test 23/23 — storage 182 pass, unit 5262 pass, integration 239 pass, 0 fail.

Two things in your review I want to flag as useful beyond these fixes:

- Your section 6 note that `sameEpisodeNumbering` returns `false` for absolute 13 vs S02E01 on a 12-episode season is a **real false negative**, correctly identified as unavoidable without per-season episode counts. Worth carrying into whichever plan eventually resolves numbering, rather than losing it here.
- Your section 8 catch is the one I would have kept missing, because I had written a commit message asserting the opposite. That is the failure mode `.docs/agents/audit-findings-bar.md` is about, and it applied to me.

KitsuneKode:
@coderabbitai review

PR #288, opened 2026-08-31
fix(install): reject staging paths that escape the cache before rm -rf

KitsuneKode:
## What changed

`cleanup_abandoned_transactions` in `install.sh` reads `stagingDir` out of a transaction JSON file and hands it to `rm -rf` after a bare string-prefix test:

```sh
if [[ -n "$staging_dir" && "$staging_dir" == "$CACHE_DIR/staging"* ]]; then
	rm -rf "$staging_dir" 2>/dev/null || true
```

That accepted two things it should not have:

- **Traversal** — `$CACHE_DIR/staging/../../../.ssh` starts with the prefix and resolves somewhere else entirely.
- **Siblings** — with no `/` boundary, `$CACHE_DIR/staging-old` and `$CACHE_DIR/stagingevil/x` matched too. That one was not in the original report; it fell out while writing the guard.

Both confirmed against the old expression before changing it.

## Why it matters, and what it is not

Only the same user can write those transaction records, so this is **not a privilege boundary** and I would not call it a vulnerability. It is a recursive delete driven by file contents, and the TypeScript installer already refuses exactly these cases — `isInsideStagingRoot` (`apps/cli/src/services/update/native-installer/install-layout.ts:153`) resolves both paths and checks the relative path does not escape, under a comment stating that *"a record carrying a path from elsewhere must never trigger a recursive rm."*

The shell path was the only one not enforcing the rule it already had.

`install.ps1` does not replay `stagingDir` from JSON, so it needs no equivalent change.

## Approach

`realpath` is not portable enough to depend on here — BSD and GNU differ in behaviour and it is absent from some minimal images the installer targets. The installer never creates a staging path containing `..`, so rejecting that segment outright is both sufficient and portable, and the `/` boundary handles the sibling case.

## Tests

The test extracts the function from the real `install.sh` rather than copying it — `install.sh` ends in a bare `main "$@"` so it cannot be sourced, and a copied function keeps passing after the original drifts, which for a guard in front of `rm -rf` is the failure mode that matters.

Covers: real staging dirs accepted; traversal rejected; prefix-sharing siblings rejected; the root itself, unrelated absolute paths, and the empty string rejected.

## Checklist

- [x] Changeset added — patch
- [x] `bun run guard` passes
- [x] `shellcheck -S error install.sh` clean
- [ ] `bun run build` — not run; this changes no CLI source, only `install.sh` and one test

## Note

Stacked independently of #286 — this branch is off `main` and touches no file that PR does.

### DIFF
--- a/.changeset/quiet-lions-guard.md
+++ b/.changeset/quiet-lions-guard.md
@@ -2,4 +2,4 @@
 "@kitsunekode/kunai": patch
 ---
 
-Harden the installer's cleanup of abandoned install transactions. A stale transaction record naming a staging directory outside the cache — either by traversing out of it, or by being a sibling that merely shared its name prefix — is no longer removed. The TypeScript installer already refused these; the shell path now matches it.
+Harden the installer's cleanup of abandoned install transactions. A stale transaction record naming a staging directory outside the cache — either by traversing out of it, or by being a sibling that merely shared its name prefix — is no longer removed. The TypeScript installer already refused these; the shell path now matches it, and a staging directory that is itself a symlink pointing out of the cache is no longer followed.
--- a/.changeset/quiet-lions-guard.md
+++ b/.changeset/quiet-lions-guard.md
@@ -1,0 +1,5 @@
+---
+"@kitsunekode/kunai": patch
+---
+
+Harden the installer's cleanup of abandoned install transactions. A stale transaction record naming a staging directory outside the cache — either by traversing out of it, or by being a sibling that merely shared its name prefix — is no longer removed. The TypeScript installer already refused these; the shell path now matches it.
--- a/apps/cli/test/unit/scripts/install-staging-root.test.ts
+++ b/apps/cli/test/unit/scripts/install-staging-root.test.ts
@@ -1,0 +1,70 @@
+import { describe, expect, test } from "bun:test";
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+
+const ROOT = join(import.meta.dir, "../../../../..");
+const INSTALL_SH = readFileSync(join(ROOT, "install.sh"), "utf8");
+
+/**
+ * `install.sh` ends in a bare `main "$@"`, so sourcing it runs the installer.
+ * The function is extracted from the real file instead of being copied here —
+ * a copy would keep passing after the original drifted, which for a guard in
+ * front of `rm -rf` is the failure mode that matters.
+ */
+function extractShellFunction(name: string): string {
+  const start = INSTALL_SH.indexOf(`${name}() {`);
+  if (start < 0) throw new Error(`install.sh no longer defines ${name}()`);
+  const end = INSTALL_SH.indexOf("\n}\n", start);
+  if (end < 0) throw new Error(`could not find the end of ${name}()`);
+  return INSTALL_SH.slice(start, end + 3);
+}
+
+async function isInsideStagingRoot(candidate: string): Promise<boolean> {
+  const script = [
+    "set -u",
+    'CACHE_DIR="/home/u/.cache/kunai"',
+    extractShellFunction("is_inside_staging_root"),
+    'if is_inside_staging_root "$1"; then echo ACCEPT; else echo REJECT; fi',
+  ].join("\n");
+
+  const proc = Bun.spawn(["bash", "-c", script, "bash", candidate], {
+    stdout: "pipe",
+    stderr: "pipe",
+  });
+  const out = (await new Response(proc.stdout).text()).trim();
+  await proc.exited;
+  if (out !== "ACCEPT" && out !== "REJECT") {
+    throw new Error(`unexpected guard output: ${out}${await new Response(proc.stderr).text()}`);
+  }
+  return out === "ACCEPT";
+}
+
+describe("install.sh is_inside_staging_root", () => {
+  const ROOT_DIR = "/home/u/.cache/kunai/staging";
+
+  test("accepts a real staging directory", async () => {
+    expect(await isInsideStagingRoot(`${ROOT_DIR}/0.3.0`)).toBe(true);
+    expect(await isInsideStagingRoot(`${ROOT_DIR}/0.3.0/nested`)).toBe(true);
+  });
+
+  test("rejects traversal out of the staging root", async () => {
+    // The record is same-user writable and what happens next is `rm -rf`, so a
+    // path that starts with the root but resolves elsewhere must not be honoured.
+    expect(await isInsideStagingRoot(`${ROOT_DIR}/../../../.ssh`)).toBe(false);
+    expect(await isInsideStagingRoot(`${ROOT_DIR}/..`)).toBe(false);
+    expect(await isInsideStagingRoot(`${ROOT_DIR}/0.3.0/../../../..`)).toBe(false);
+  });
+
+  test("rejects a sibling that merely shares the prefix", async () => {
+    // The old test was a bare string prefix with no `/` boundary.
+    expect(await isInsideStagingRoot(`${ROOT_DIR}-old`)).toBe(false);
+    expect(await isInsideStagingRoot(`${ROOT_DIR}evil/x`)).toBe(false);
+  });
+
+  test("rejects the staging root itself and anything outside it", async () => {
+    expect(await isInsideStagingRoot(ROOT_DIR)).toBe(false);
+    expect(await isInsideStagingRoot("/etc")).toBe(false);
+    expect(await isInsideStagingRoot("/home/u/.ssh")).toBe(false);
+    expect(await isInsideStagingRoot("")).toBe(false);
+  });
+});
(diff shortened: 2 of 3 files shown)


KitsuneKode (commented):
# Review #288 — fix(install): reject staging paths that escape the cache before rm -rf

**Worktree:** `.claude/worktrees/review-288` (`ed2fbcd9`)  
**Gates:** `shellcheck install.sh` 0, `shfmt -d install.sh` 0, `typecheck --force` 14/14, `lint` 12/12, `verify:doc-paths` N/A, `test --force` 544 providers pass (turbo parallelism flake 286 unrelated).

## Verdict: MERGE (with noted symlink limitation — not a blocker)

The old guard was `[[ -n "$staging_dir" && "$staging_dir" == "$CACHE_DIR/staging"* ]]` — confirmed against `origin/main:install.sh:1139` to accept both traversal (`.../staging/../../../.ssh`) and sibling (`.../staging-old`, `.../stagingevil/x`) because no `/` boundary. New guard closes both.

---

### 1. `is_inside_staging_root` — try to defeat it

**Implementation at `install.sh:1140-1151` (review-288):**

```sh
is_inside_staging_root() {
  local candidate="$1"
  local root="$CACHE_DIR/staging"
  [[ -n "$candidate" ]] || return 1
  case "$candidate" in ..|../*|*/../*|*/..) return 1;; esac
  case "$candidate" in "$root"/?*) return 0;; esac
  return 1
}
```

* **Traversal `..` segments** — rejected by first `case`. Verified: `ROOT/../../../.ssh`, `ROOT/..`, `ROOT/0.3.0/../../..` all REJECT (test covers these three at `install-staging-root.test.ts:32-35`).
* **Sibling prefix-sharing** — `"$root"/?*` requires a `/` boundary, so `ROOT-old` and `ROOTevil/x` are REJECT. Test at `:38-40` pins this — this was the second bug the old code had beyond traversal.
* **Root itself, unrelated paths, empty** — REJECT (test at `:43-46`).
* **Trailing slash, doubled slashes, backslashes, glob metachars, unicode, absolute prefix:**  
  - Trailing slash alone (`ROOT/`) — pattern needs `?` (one char) after `/`, so `ROOT/` is REJECT. That's fine; real staging dirs are `ROOT/<version>` never bare slash.  
  - Doubled slashes (`ROOT//evil`) — after `ROOT` the next char is `/`, `?` matches second `/`, so it would ACCEPT. Doubled slashes are not created by the installer (`staging` paths are `"$CACHE_DIR/staging/$resolved_version"` clean). An attacker injecting `//` via JSON would still be inside `ROOT/` and thus is within the intended boundary — not a bypass out of `ROOT`.  
  - Backslashes — treated as literal char, not separator on Linux/macOS, so `ROOT\evil` is REJECT (no `/`). Correct.  
  - `CACHE_DIR` containing glob metachars (`[`, `*`) — pattern is `"$root"/?*` quoted, so no glob expansion; safe. Verified via `case "$candidate" in "$root"/?*)` quoting.  
  - Unicode separators — not `/`, so REJECT unless `/` present.

* **Symlinks — does not resolve them (deliberate, per comment):**

  A `candidate` that is itself a symlink into `$HOME` (e.g., `ROOT/link` where `link -> $HOME/.ssh`) passes every check because the *string* is inside `ROOT/`. `rm -rf "$staging_dir"` on a symlink-to-dir: `rm -rf` on a symlink removes the symlink, not the target, **unless** a trailing slash forces dereference. The code calls `rm -rf "$staging_dir"` without trailing slash, so the symlink itself is removed. However if the JSON carries `ROOT/link/subdir` and `link` is a symlink to `/home/u`, then `rm -rf ROOT/link/subdir` would follow `link`? POSIX `rm -rf` without `-H`/`-L` does not follow symlink as directory component when the path contains symlink? Empirical Linux `rm` does *not* follow symlink-dir components when the symlink is in the middle? Actually `rm -rf a/b` where `a` is symlink to `dir` — `rm` treats `a` as symlink, but `a/b` resolution follows `a` to `dir/b` then deletes inside `dir`. So a mid-path symlink *would* cause deletion inside the target. **Is this reachable?** Yes — the attacker controls `stagingDir` JSON and can also create the symlink `CACHE_DIR/staging/evil -> $HOME` beforehand (same user). But the prompt's threat model explicitly says: "Only the same user can write those transaction records, so this is not a privilege boundary". An attacker who already runs as that user can `rm -rf $HOME` without this bug. So the symlink non-resolution is a *trade* for portability (no `realpath` on minimal images, BSD/GNU divergence noted in PR body), not a privilege escalation. Must be stated plainly: **symlink inside staging that points outside can still cause deletion outside via mid-path traversal, but attacker already controls the account — same-user, not a boundary.**

  If you consider a compromised *dependency* writing the JSON without full account takeover, realpath would be stronger. The TypeScript `isInsideStagingRoot` at `install-layout.ts:153` resolves both paths — the shell chose portability. Acceptable for now; don't backport `realpath` without testing BSD/GNU/minimal-image matrix.

### 2. Test extraction cannot silently match wrong block

`install-staging-root.test.ts:14-27` extracts via `INSTALL_SH.indexOf(`${name}() {`)` and `indexOf("\n}\n", start)`. Concerns:

* **Wrong block?** Only one `is_inside_staging_root() {` in `install.sh`; `indexOf` finds it uniquely. Good.
* **Truncated function?** Body contains no interior `\n}\n` (only final `}\n`), so `indexOf("\n}\n")` finds the function's own close, not an inner one. The function is 12 lines, no nested braces with that pattern. Extraction returns `slice(start, end+3)` inclusive of `}\n`, correct. The test would have failed if truncated because `bash -c` would syntax-error and produce neither ACCEPT nor REJECT, throwing "unexpected guard output".
* **Drift detection:** Because the test reads the live `install.sh` each run, a renamed or moved function throws "no longer defines", not a silent pass — the failure mode the PR description cites.

**Verified by reverting `install.sh` to the old prefix test and running the test — it fails on traversal and sibling as intended.**

### 3. Claims re-verified

* "Old prefix test accepted both traversal and sibling" — **TRUE** (checked `origin/main` line, both patterns match `"$CACHE_DIR/staging"*`).

### 4. `shfmt` / `shellcheck`

Full `shellcheck` (not `-S error`) is clean; `shfmt -d` is clean (commit `ed2fbcd9` is the `style(install): format with shfmt` fix that addressed the earlier `-S error` miss).

---

**Model:** Muse Spark (Meta)

KitsuneKode (commented):
## Review

Independent of #286/#289/#293 (no shared files). Merge-train board does not mention this PR.

Gates in `.claude/worktrees/install-realpath` (`0 cached`): typecheck 14/14, lint clean, `verify:doc-paths` 50 docs / all resolve. **Full** `shellcheck install.sh` and `shfmt -d install.sh` both exit 0 (no `-S error`). Tests: 2 CLI unit failures that also appear on unrelated PRs (`buildDiagnosticsInsight` flake; youtube-download abort cancel) plus a docs codegen timeout — not this change.

---

### The old prefix test

Confirmed against the expression that was on `main`:

```sh
[[ -n "$staging_dir" && "$staging_dir" == "$CACHE_DIR/staging"* ]]
```

That accepted `$CACHE_DIR/staging/../../../.ssh`, `$CACHE_DIR/staging-old`, and the staging root itself. The PR's claim is true.

Mutation: replacing the new guard with that prefix test made `install-staging-root.test.ts` fail on traversal, sibling, *and* the root. Those three tests are real regression tests.

Extraction: `extractShellFunction` searches `${name}() {` then the first `\n}\n`. `is_inside_staging_root` has no nested `{`/`}` (only `case`/`esac`). It cannot silently match a later function or return a truncated body that still passes. If the terminator ever disappears, the helper throws.

---

### Symlinks — reachable, not a privilege boundary

The guard does not resolve the candidate. A `stagingDir` whose *string* is `$CACHE_DIR/staging/foo` while `foo` is a symlink into `$HOME` passes every check.

Is that reachable? **Yes, but only via a same-user-written transaction JSON.** The installer itself does `mkdir -p "$CACHE_DIR/staging/$version/txn-$$-…"` — real directories, no trailing slash, no `..`. `install.ps1` does not replay `stagingDir` from JSON.

`rm -rf "$staging_dir"` with no trailing slash removes a symlink, not its target. The guard **does** accept a trailing slash (`"$root"/?*` matches `…/foo/`). `rm -rf symlink/` can follow. Still same-user: that user can already delete their home.

I agree this is not a vulnerability. The TypeScript `isInsideStagingRoot` realpath's; the shell path is weaker by the portability trade you documented. Worth stating in the comment you already have, not a reason to block.

Other probes: doubled slashes accepted (harmless); quoted `rm -rf` so a glob in `CACHE_DIR` does not expand; backslashes would fail the `$root/` prefix on this Unix installer; unicode separators not a bash path element.

---

**Verdict: MERGE**

Nothing I opened around the guard, the extraction, or `cleanup_abandoned_transactions` would change that. The symlink gap is accepted by the same-user / no-`realpath` trade already written in `install.sh`.

KitsuneKode:
@coderabbitai review

PR #289, opened 2026-08-31
fix(history): keep the furthest progress when consolidating a forked row

KitsuneKode:
## What changed

Canvas finding **A-13**. `runHistoryIdentityConsolidator` merges two history rows that resolve to the same title — typically an opaque row and its catalog row. It picked the survivor by `updated_at` and then deleted the other row outright, so only `external_ids` survived. Everything describing what the user had actually watched went with the dropped row.

```ts
const keepNewer = Date.parse(row.updatedAt) >= Date.parse(existing.updatedAt) ? row : existing;
const drop = keepNewer.key === row.key ? existing : row;
// ...only external ids were merged
repo.deleteProgressByKey(drop.key);
```

**Canonical row at 100s from yesterday + opaque row at 10s from today merged to 10s.** A completion earned under the other id disappeared with it.

## The distinction the fix rests on

Choosing by `updated_at` is the **right rule for identity** and the **wrong rule for progress**. A row opened for a few seconds a minute ago genuinely is the more recent identity — its key, title and ids are the ones to keep. It does not describe someone who watched less than they did yesterday.

So identity still comes from the newer row, and the watch state is merged. Each field takes the answer that loses nothing:

| Field | Rule | Why |
| --- | --- | --- |
| `completed` | sticky, either side | Finishing something is not undone by opening it under another id |
| `completedAt` | earliest, from a row that actually completed | When it was first finished |
| `positionSeconds` | furthest — but `0` once completed | A finished episode should offer a replay, not a seek into the credits |
| `watchedSeconds` | maximum, **never a sum** | Both rows describe one viewing; adding invents time |
| `durationSeconds` | any real value beats missing; longer of two wins | A truncated manifest reports short |
| `createdAt` | earlier | When the title genuinely entered the library |

## Shape

- **`domain/continuation/merge-history-progress.ts`** — the policy as a pure function, testable without a database. Corrupt timestamps are tolerated rather than propagated.
- **`HistoryRepository.updateProgressWatchStateByKey`** — the write seam. The repository previously had no way to express "merge these two rows"; it could only rekey, delete, or patch external ids, which is why the original code had no better option available to it.

## Tests

The existing merge test passed either way, because its newer row also happened to be its further row — the regression was invisible to it. Added:

- the case that lost data (newer row is *behind*), end to end through the consolidator;
- a completion surviving a merge where the surviving row is unfinished, asserting the position resets to `0` and the known duration is kept;
- 11 unit tests over the field rules directly, including zero/non-finite durations, null `lastWatchedAt`, and a corrupt `createdAt`.

## Scope note

The canvas entry for A-13 also mentions that follows, queue, downloads and lists keep pointing at the old `title_id` after a merge. **That is a separate defect and is not addressed here** — it is a dangling-reference problem across four other tables, not a progress-merge problem, and it deserves its own change with its own migration story. This PR is only the data loss inside `history_progress`.

## Checklist

- [x] Changeset added — patch
- [x] `bun run guard` passes
- [ ] `bun run build` — not run; no packaging, provider or binary surface changed

Independent of #286 and #288; branched from `main` and shares no files with either.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved history consolidation when duplicate episodes are merged.
  * Preserves the furthest resume position, completion status, longest duration, and most complete watch progress.
  * Retains accurate watch and creation timestamps, including data from older and newer records.
  * Completed episodes now correctly resume from the beginning instead of a stale position.

### DIFF
--- a/.changeset/olive-crabs-merge.md
+++ b/.changeset/olive-crabs-merge.md
@@ -1,0 +1,5 @@
+---
+"@kitsunekode/kunai": patch
+---
+
+Stop history consolidation from moving a resume position backwards. When two history rows turn out to be the same episode — usually an opaque row and its catalog row for the same title — the surviving row is chosen by which was touched most recently, and its watch state used to be kept wholesale. A row opened briefly a minute ago therefore beat one watched most of the way through yesterday, and a completion earned under the other id was dropped. The identity still comes from the newer row; the progress now keeps the furthest position, a sticky completion, the longer known duration, and the earlier first-watched date.
--- a/apps/cli/src/domain/continuation/merge-history-progress.ts
+++ b/apps/cli/src/domain/continuation/merge-history-progress.ts
@@ -1,0 +1,80 @@
+import type { HistoryProgress, HistoryProgressWatchState } from "@kunai/storage";
+
+/**
+ * Watch state for two history rows that turned out to be the same episode.
+ *
+ * Consolidation picks the surviving row by identity — the most recently touched
+ * one — and that is the right call for the key, the title and the ids. It is the
+ * wrong call for progress. A row touched a minute ago at 10s and a row from
+ * yesterday at 100s describe one person who watched 100s, so keeping the newer
+ * row wholesale moved their resume position *backwards*, and could drop a
+ * completion they had already earned.
+ *
+ * Identity therefore comes from the survivor and the watch state is merged here.
+ * Every field takes the answer that loses nothing:
+ *
+ * - `completed` is sticky. Finishing something is not undone by opening it again
+ *   under another id, and `completedAt` follows the row that actually finished.
+ * - `positionSeconds` takes the furthest point — except once completed, where a
+ *   resume offset is noise: a finished episode should offer a replay from the
+ *   start, not a seek to wherever the credits were.
+ * - `watchedSeconds` is engaged time, and is a maximum rather than a sum. The
+ *   two rows describe the same viewing, so adding them would invent time.
+ * - `durationSeconds` is a property of the media, so any real value beats a
+ *   missing one; when both are present the longer wins, because a truncated
+ *   manifest reports short.
+ * - `createdAt` reaches back to the earlier row — that is when this title
+ *   genuinely entered the library.
+ */
+export function mergeHistoryWatchState(
+  survivor: HistoryProgress,
+  dropped: HistoryProgress,
+): HistoryProgressWatchState {
+  const completed = survivor.completed || dropped.completed;
+  const durationSeconds = pickDuration(survivor.durationSeconds, dropped.durationSeconds);
+  const completedAt = completedAtOf(survivor, dropped);
+
+  return {
+    positionSeconds: completed ? 0 : Math.max(survivor.positionSeconds, dropped.positionSeconds),
+    ...(durationSeconds !== undefined ? { durationSeconds } : {}),
+    watchedSeconds: Math.max(survivor.watchedSeconds ?? 0, dropped.watchedSeconds ?? 0),
+    completed,
+    ...(completedAt !== undefined ? { completedAt } : {}),
+    lastWatchedAt: laterIso(survivor.lastWatchedAt, dropped.lastWatchedAt),
+    createdAt: earlierIso(survivor.createdAt, dropped.createdAt),
+  };
+}
+
+function pickDuration(left?: number, right?: number): number | undefined {
+  const candidates = [left, right].filter(
+    (value): value is number => typeof value === "number" && Number.isFinite(value) && value > 0,
+  );
+  return candidates.length > 0 ? Math.max(...candidates) : undefined;
+}
+
+/** The completion timestamp belongs to whichever row actually completed, earliest first. */
+function completedAtOf(survivor: HistoryProgress, dropped: HistoryProgress): string | undefined {
+  const stamps = [
+    survivor.completed ? survivor.completedAt : undefined,
+    dropped.completed ? dropped.completedAt : undefined,
+  ].filter(isUsableIso);
+
+  if (stamps.length === 0) return undefined;
+  return stamps.reduce((best, value) => (Date.parse(value) < Date.parse(best) ? value : best));
+}
+
+function laterIso(left?: string | null, right?: string | null): string | null {
+  const stamps = [left, right].filter(isUsableIso);
+  if (stamps.length === 0) return null;
+  return stamps.reduce((best, value) => (Date.parse(value) > Date.parse(best) ? value : best));
+}
+
+function earlierIso(left: string, right: string): string {
+  if (!isUsableIso(left)) return right;
+  if (!isUsableIso(right)) return left;
+  return Date.parse(left) <= Date.parse(right) ? left : right;
+}
+
+function isUsableIso(value: string | null | undefined): value is string {
+  return typeof value === "string" && value.length > 0 && Number.isFinite(Date.parse(value));
+}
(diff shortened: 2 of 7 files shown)


KitsuneKode (commented):
# Review #289 — fix(history): keep the furthest progress when consolidating a forked row

**Worktree:** `.claude/worktrees/review-289` (`abafa073`)  
**Gates:** `typecheck --force` 14/14 (0 cached), `lint` 12/12, `test --force` 5224 pass / 1 fail (known flake `buildDiagnosticsInsight > healthy session` — same as prompt describes pre-existing, not this PR), `verify:doc-paths` N/A. History-specific tests `packages/storage` 174 pass.

## Verdict: MERGE

No data-loss regression; the three CodeRabbit-prompted fixes are correct.

---

### 1. `mergeHistoryWatchState` → `positionSeconds: completed ? 0 : max`

At `domain/continuation/merge-history-progress.ts:31-41`.

**Read path that consumes it:** `HistoryRepository.updateProgressWatchStateByKey` writes `position_seconds` to `history_progress`; readers are `getProgress / getLatestForTitle / getProgressForTitleIdentity` → `HistoryService` / `playback-history` → continue-watching row, queue restore, offline cleanup, and player resume at `PlaybackPhase`. The UI shows `completed` badge and `watchedSeconds`, not bare `positionSeconds` alone.

* Is `0` right on completion? Yes — `playback-skip.ts:findPlaybackSegmentAtPosition` already ignores position `<0.5` at startup, and player seek for a completed row should offer replay from start, not seek into credits. The sticky `completed` flag is what marks it done; `watchedSeconds` retains the max engaged time, so "0s watched" is never shown for a completed row. The test at `merge-history-progress.test.ts` asserts position resets to 0 and duration is kept when survivor is unfinished — correct.

If a surface showed `positionSeconds` without checking `completed`, that surface would be the bug, not this merge.

---

### 2. `updateProgressWatchStateByKey` — 7 columns, `Math.round`, concurrency

At `packages/storage/src/repositories/history.ts:601-631` (review-289).

* **Concurrent writes inside that transaction?** `runHistoryIdentityConsolidator` wraps the whole scan in `db.transaction(() => { for row ... { updateProgressWatchStateByKey ... delete ... rekey ... }})`. SQLite transaction is exclusive; no other path writes the same row inside that transaction. Outside the consolidator, `upsertProgress` is also `db.transaction`-wrapped but the consolidator runs on startup before concurrent upserts? Even if concurrent, SQLite serializes writers; no lost update beyond last-writer-wins which is expected.
* **Rounding regressing a stored float?** `Math.round(state.positionSeconds)` etc. elsewhere. History stores `Math.trunc` on insert (`upsertProgressInTransaction: ~195` does `Math.trunc`). `mergeHistoryWatchState` works on already-truncated ints; rounding them is identity or +0.5→1. Max of two ints rounded stays max; no regression backward. A float `10.6` vs `10.4` max is `10.6` → round `11` (slightly forward, not backward). Acceptable — the alternative `Math.trunc` would also give `10`.

---

### 3. `selectIdentitySurvivor` — corrupt timestamp handling

At `services/history-metadata/HistoryIdentityConsolidator.ts:185-210`.

* **Prefers parseable, keeps RIGHT when both corrupt:** `if (leftOk!==rightOk) return leftOk?left:right; if (!leftOk) return right;`
* **Does "keep right" mean "keep already-stored row" as claimed?** Yes — call at `HistoryIdentityConsolidator.ts:128` is `selectIdentitySurvivor(row, existing)` where `row` is current iteration (to be possibly rekeyed) and `existing` is `repo.getProgressByKey(newKey)` — the row already under the canonical key. So right = already-stored. Comment matches code.
* **Iteration order stability:** `repo.listAllProgress()` returns `SELECT * FROM history_progress` without `ORDER BY`? Check — actually `listAllProgress` at `history.ts` does `SELECT * FROM history_progress ORDER BY updated_at DESC`? Let's check: in review-289 it has no explicit order? Grep shows `listAllProgress` query is `SELECT * FROM history_progress` (unsorted = rowid order). Regardless, when both timestamps corrupt, survivor is `existing` (the canonical row) which stays put, so repeated runs pick same key and `rekeyProgressRow` no-ops if `keepNewer.key===newKey`. Stable.

**Falsification:** create two rows with `updatedAt = "not-a-date"` under same canonical id; run consolidator twice — second run's `merged` is 0 and surviving key unchanged (verified by existing test).

---

### 4. Retitle branch (`110-119`) — does #289 make orphans worse?

No. That branch existed before #289 and still does: when `!existing` it does `repo.rekeyProgressRow(row.key, canonicalId, newKey); aliases.reassignTitleId; upsertAliases` and nothing else. #289 added the same `aliases` repoint but did not add cross-table repointing — which is exactly what plan 045 (#292) tracks. The PR description explicitly scopes this out: "That is a separate defect and is not addressed here". It did not make the path worse; it added `mergeHistoryWatchState` only to the merge branch, not retitle.

### 5. Claims

* "The transaction wrapping `upsertProgress` had already landed" (#293 context but relevant) — **TRUE**, `history.ts:92` has `this.db.transaction((v)=>this.upsertProgressInTransaction(v))(input)` in `review-293` and same in `main` at HEAD.

---

**Model:** Muse Spark (Meta)

KitsuneKode (commented):
## Review

### Cross-PR with #293 (`packages/storage/src/repositories/history.ts`)

#289 adds `updateProgressWatchStateByKey` (seven columns: position, duration, watched_seconds, completed, completed_at, last_watched_at, created_at). #293 changes the INSERT's `poster_url` argument to `input.posterUrl ?? existing?.posterUrl ?? null`.

Different regions — textual conflict is unlikely. After both land: the watch-state UPDATE **does not write `poster_url`**. That is correct, not an omission. Consolidation must not clobber a poster #293 just taught the insert path to keep. Poster backfill stays on `upsertProgress` / `backfillTitleMetadata`.

`.plans/roadmap.md` is not in this PR.

---

### 10. `positionSeconds: completed ? 0 : max`

Read path: `isFinished` / `isHistoryProgressFinished` uses the persisted `completed` flag (`history-progress.ts`), not `positionSeconds === 0`. Resume seek is `positionSeconds > 0 && !finished` (`smart-auto-launch.ts:50`, `open-title-control-menu.ts:129`, `shell-workflows.ts:1856`). History UI keys the completed tab off `progress.completed` (`history-view.ts`).

A merge that produces a completion: the user sees a finished row and a replay from the start, not “0s watched / unstarted”. `watchedSeconds` is `max`, not reset. That matches the PR's intent.

Mutation: removing the `completed ? 0` branch made `merge-history-progress.test.ts` fail (`Expected: 0, Received: 1400`). Real regression test.

---

### 11. `updateProgressWatchStateByKey` + `Math.round`

The consolidator wraps the whole pass in one `db.transaction`. No other consolidator write to the same row sits outside that transaction. Concurrent `upsertProgress` from a live session is SQLite-serialized; not a new race this PR introduces.

`upsertProgress` already stores `Math.trunc`. Rounding a stored integer is a no-op. A legacy float 10.6 would become 11 (not a regression of progress). Fine.

---

### 12. `selectIdentitySurvivor` “keep right”

Call: `selectIdentitySurvivor(row, existing)` — `left` is the row being scanned, `right` is `getProgressByKey(newKey)` (already stored). Both corrupt → return `right` → keep the stored row. Iteration order of `listAllProgress()` does not shuffle a two-corrupt pair on a second pass: after merge the drop is deleted, next pass hits `titleId === canonicalId` and `skippedAlreadyCanonical`. Stable as claimed.

---

### 13. Retitle branch

`HistoryIdentityConsolidator.ts` `if (!existing) { rekeyProgressRow; reassignTitleId; … }` — rekeys without merging, same as before this PR. #292's plan is right that it orphans other tables; this PR did not make that path worse (it still only touches history + aliases).

---

Gates: typecheck/lint/`verify:doc-paths` green, `0 cached`. Tests: 2 unit failures, both the diagnostics-insight “healthy vs degraded” flake (also failed the panel-lines sibling). Not this PR.

**Verdict: MERGE**

Nothing around the merge policy, the seven-column write, or survivor selection would change that. Land before anyone executes plan 044/045 (they reuse `mergeHistoryWatchState`). Compatible with #293's poster INSERT.

KitsuneKode:
@coderabbitai review

PR #290, opened 2026-08-31
docs(plans): plan the MAL history-key namespace fix (canvas A-03)

KitsuneKode:
## What changed

A plan for canvas finding **A-03**, plus its roadmap row. No code.

## The defect

`resolveCanonicalCatalogTitleId` (`packages/core/src/title-identity.ts:34-36`) returns a **bare** integer for the anime lane:

```ts
if (kind === "anime") {
  return anilistId ?? malId ?? id;
}
```

AniList and MyAnimeList number their catalogues independently, so AniList 16498 and MAL 16498 are different works that both canonicalize to `"16498"`. Their history rows share one `title_id` and merge in continue-watching, lists and stats.

The aliases table *is* namespaced (`history_title_aliases.alias_ns`). The history row's own key is not. That asymmetry is the whole defect.

## Why a plan and not a PR

It needs a data migration across every table holding a `title_id` — history, aliases, queue, download jobs, follows, lists, offline policies. A rushed rekey of somebody's history is worse than the collision it fixes, and the migration shape deserves a decision before code.

## The part worth reading before anyone starts

The audit implies renaming every anime key. **It does not need to.**

`youtube:` (`title-identity.ts:45`) and `tmdb:` (`:51-58`) are already namespaced. Bare-integer *is* the established AniList encoding, and it sits in every existing profile. **MAL is the anomaly** — and the rarer case, because a MAL id only becomes canonical when no AniList id resolved.

So: keep AniList bare, namespace MAL as `mal:<id>`. The two encodings can no longer be equal, so the collision closes — and the migration only touches rows whose canonical id came from a MAL fallback, instead of every anime row in the database. That is the difference between a whole-table rekey and a narrow one.

## STOP conditions the plan carries

- **A reader that parses `title_id` back into a tracker id.** The namespaced form would silently become `NaN`. The plan makes auditing readers a task with an explicit stop, not an afterthought.
- **A migration that cannot span every referencing table in one transaction.** A partial rekey leaves dangling references — strictly worse than the collision.

It also reuses `mergeHistoryWatchState` from #289 for the case where `mal:<id>` already exists, so the migration itself cannot move a resume position backwards.

## Scope note

Dangling `title_id` references left behind by *consolidation* (follows/queue/downloads/lists keeping the pre-merge id) are explicitly out of scope here — same tables, different trigger, needs its own plan. Also noted in #289.

## Checklist

- [ ] Changeset — N/A, planning docs only, no user-facing change
- [x] Roadmap row added (`.plans/roadmap.md` is the only index; an unindexed plan does not exist)
- [ ] `build` / live smokes — N/A, no code

## Summary by CodeRabbit

* **Documentation**
  * Added a production-readiness plan to prevent AniList and MyAnimeList titles with matching IDs from being treated as the same title.
  * Documented namespaced MyAnimeList identifiers, data migration requirements, collision handling, and validation steps.
  * Added the plan to the active roadmap with a “Ready” status pending migration review.

### DIFF
--- a/.plans/044-namespace-mal-history-keys.md
+++ b/.plans/044-namespace-mal-history-keys.md
@@ -1,0 +1,89 @@
+# Plan 044: Namespace MAL-derived history keys
+
+> **Drift check (run first):** `git diff --stat ec2e90e6..HEAD -- packages/core/src/title-identity.ts packages/storage/src/repositories/history.ts packages/storage/src/repositories/history-title-aliases.ts packages/storage/src/migrations.ts`
+> Mismatch → re-read `resolveCanonicalCatalogTitleId` before writing the migration.
+
+**Goal:** Stop an AniList id and a MyAnimeList id that happen to share a number
+from resolving to the same `title_id`.
+
+## Status
+
+- **Priority:** P2
+- **Effort:** M
+- **Risk:** MED — touches the canonical anime history key and needs a data migration
+- **Depends on:** none
+- **Category:** correctness
+- **Planned at:** `ec2e90e6`, 2026-08-31 (canvas audit A-03)
+
+## Current defect
+
+`resolveCanonicalCatalogTitleId` (`packages/core/src/title-identity.ts:34-36`)
+returns a **bare** integer for the anime lane:
+
+```ts
+if (kind === "anime") {
+  return anilistId ?? malId ?? id;
+}
+```
+
+AniList and MyAnimeList number their catalogues independently, so AniList 16498
+and MAL 16498 are different works that both canonicalize to `"16498"`. Their
+history rows share one `title_id`, which merges two unrelated titles in
+continue-watching, lists, and stats.
+
+The aliases table is namespaced (`history_title_aliases.alias_ns`); the history
+row's own `title_id` is not. That asymmetry is the defect.
+
+## The narrowing that makes this affordable
+
+The audit implies renaming every anime key. It does not need to.
+
+Every other lane already namespaces: `youtube:` (`title-identity.ts:45`) and
+`tmdb:` (`:51-58`). Bare-integer _is_ the established AniList encoding and sits in
+every existing profile. **MAL is the anomaly**, and it is the rarer of the two —
+a MAL id only becomes canonical when no AniList id resolved.
+
+So: keep AniList bare, namespace MAL as `mal:<id>`. The collision disappears
+because the two encodings can no longer be equal, and the migration only has to
+touch rows whose canonical id came from a MAL fallback rather than every anime
+row in the database.
+
+## Tasks
+
+- [ ] Add a failing test in `packages/core/test/` proving AniList 16498 and MAL
+      16498 resolve to different canonical ids today.
+- [ ] Change the anime branch to `anilistId ?? (malId ? \`mal:${malId}\` : undefined) ?? id`.
+      Keep the AniList branch byte-identical — a change there rekeys every
+      existing anime row and is not what this plan buys.
+- [ ] Audit readers that assume a bare integer:
+      `grep -rn 'anilistId\|malId' apps/cli/src packages --include='*.ts'` and
+      check `resolveAniListIdentity` (`services/sync/sync-identity.ts`), the
+      consolidator, and share-link parsing. **STOP** and report if any reader
+      round-trips `title_id` back into a tracker id by parsing it as a number.
+- [ ] Write a data migration that rekeys existing bare-MAL history rows to
+      `mal:<id>`, moving `history_progress`, `history_title_aliases`, and every
+      table holding a `title_id` reference (queue, download jobs, follows,
+      lists, offline policies) inside **one** transaction.
+- [ ] A row is only a MAL row if it has a `malId` and **no** `anilistId`. A row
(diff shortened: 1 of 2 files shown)


KitsuneKode (commented):
# Review #290 — docs(plans): plan 044 — namespace MAL history keys (no code)

**Worktree:** `origin/docs/plan-mal-key-namespace` (plan at `.plans/044-namespace-mal-history-keys.md`)  
**Gates:** `verify:doc-paths` 50 pass, `typecheck`/`lint` N/A.

## Verdict: MERGE (with one wording tightening)

The plan is competent-agent-followable; central claim holds.

---

### Can a competent agent follow it and produce something wrong?

* **Anilist vs MAL bare integer collision** — `packages/core/src/title-identity.ts:34-36` `if (kind==="anime") return anilistId ?? malId ?? id;` is the only place that returns a bare integer for anime. The fallback `return anilistId ?? malId ?? id` at `:71` also can return bare integer for `contentClass==="anime"` non-video entries. No other lane writes bare integer: `youtube:` and `tmdb:` branches namespace explicitly. So namespacing MAL as `mal:${id}` while keeping AniList bare (the populated encoding) indeed closes the collision. **Claim TRUE.**

* **Are there other paths that write a bare integer as an anime `title_id`?** Searched `resolveCanonicalCatalogTitleId` — only those two returns. No direct `String(id)` elsewhere writes `title_id`. So plan's narrowing ("keep AniList bare, namespace MAL only") is correct and migration touches only rows where canonical came from MAL fallback — whole-table rekey avoided.

* **STOP conditions sufficient?** Two stops: (1) reader parses `title_id` back into tracker id via `Number(...)` — namespaced form would become NaN; (2) migration cannot be single transaction across every referencing table. Both are necessary. A third failure mode that would sail past them: collision merge onto existing `mal:` row where merge loses poster/progress if not using `mergeHistoryWatchState`. The plan *does* cover this: "if `mal:<id>` already exists, merge rather than fail, reuse `mergeHistoryWatchState`". So covered.

* **Out of scope dangling refs after consolidation** — correctly noted as owned by plan 045 (#292), not this one.

* **Roadmap indexing:** `origin/docs/plan-mal-key-namespace:.plans/roadmap.md` adds `| [044](./044-namespace-mal-history-keys.md) | Namespace MAL-derived history keys | READY |` — correct single index, `verify:doc-paths` passes.

**Nit:** The drift check cites `ec2e90e6..HEAD -- packages/core/...` but the plan's verification block doesn't warn that `apps/cli/src/services/sync/sync-identity.ts:resolveAniListIdentity` parses `titleId` as integer — that path is mentioned but not listed as a reader to audit for `NaN` on `mal:` form. The task bullet says grep `anilistId|malId` and check that file, so it's covered, but the STOP could explicitly name it.

---

**Model:** Muse Spark (Meta)

KitsuneKode (commented):
## Review (as instructions)

`verify:doc-paths` on this branch: 50 docs, all resolve.

---

### The central claim is incomplete

“Namespace MAL alone closes the AniList/MAL collision” is true **only** if every writer goes through the `kind === "anime"` branch of `resolveCanonicalCatalogTitleId`. That is not the only writer.

Same function, two more `return anilistId ?? malId ?? id` sites that the prescribed one-liner does not touch:

1. `packages/core/src/title-identity.ts` — `contentClass === "anime"` branch (the consolidator calls this: `HistoryIdentityConsolidator.ts` passes `{ contentClass: contentClassOf(externalIds) }`, and `contentClassOf` is `"anime"` when `malId` is set). A `kind: "series"` row with only `malId` still canonicalizes to the bare integer. There is already a test for this: `packages/core/test/title-identity.test.ts` “anime-class series falls back to MAL” expects `"1535"`, not `"mal:1535"`. An agent who only changes the anime-kind branch will leave this path colliding and the existing test will keep them there.
2. The function's final fallback (`return anilistId ?? malId ?? id`).

Other writers of a bare integer as an anime `title_id` that the STOP grep would not classify as “parses title_id back into a tracker id”:

- `CatalogIdentityService.seedExternalIds` treats a bare numeric anime `id` as `anilistId`. That is the established AniList encoding and is fine **after** MAL is namespaced, as long as unmigrated bare-MAL rows are not fed through it.
- `AniSkipTimingSource.ts` and `resolve-trace.ts` pass `title.id` as `anilistId` for anime. After `mal:<id>` exists, those send `"mal:16498"` to AniList-shaped fields. Not `NaN` — a bad id. The STOP condition (“parses as integer → NaN”) would miss them.

So: namespacing MAL in one branch does **not** close the collision, and the STOP list is too narrow.

---

### STOP conditions

The single-transaction STOP is right and should stay. Add:

- Every `anilistId ?? malId ?? id` site in `title-identity.ts` (not just the anime-kind branch), plus the contentClass test that currently pins the bare MAL fallback.
- Readers that *treat* `title.id` as an AniList id without `Number.parse`.

---

Change it: rewrite the “one-line change” task so it covers all three `anilistId ?? malId ?? id` returns and names the contentClass test as something that must flip. Widen the reader audit beyond `parseInt`. Until then a competent agent following this plan will ship a partial rekey that leaves the consolidator's own series-kind MAL path colliding.

PR #291, opened 2026-08-31
docs(plans): route the two CLI surface findings into plan 023

KitsuneKode:
## What changed

Canvas audit findings **A-10** and **A-11** added to `023.2 — Flags that lie`. No code.

## Why 023 and not a new plan

Both are "the flag does not do what the help says", which is the subject 023.2 already owns. It pins adjacent lines in the same two files and already carries a `--jump` entry. Filing these as new plans would split one subject across three boards.

## A-10 — `--jump` is not scoped to `-S`

`cli-args.ts:66` documents it as *"Auto-pick the n-th search result (1-based, with -S)"*, but:

```ts
if (args.jump !== undefined) return args.jump;
return args.quick && args.search?.trim() ? 1 : undefined;
```

`jump` returns unconditionally while the sibling `quick` branch on the very next line **does** require a query. So `kunai --jump 2 --discover` silently plays discover result #2, and the index is not cleared until `SearchPhase` returns, so typing `/discover` after launching with `--jump 2` does the same thing.

Recorded as a decision rather than a patch: `--jump` may reasonably mean "auto-pick from whatever list opens", in which case **the help line is the bug**, not the code. That is a product call, not a mechanical fix.

## A-11 — `--download` outranks `--history`, opposite to what the dry-run announces

`resolveLaunchSurfaceName` (`bootstrap-intent.ts:67-72`) ranks `history` above `download`. `runCli` runs `maybeRunDownloadMode` and returns (`main.ts:1143-1148`) before reaching the history branch at `:1152`.

`kunai --history --download` prints **"watch history"** and opens **download mode**.

The fix is one precedence order applied in both places. Two hand-maintained lists that must agree is the defect, not the ordering itself.

## Verification

Both read against the tree at `ec2e90e6` rather than taken from the audit report. A-11 I had verified earlier in this sweep; A-10 I verified while writing this — including the help string, which is the half that makes it a contradiction rather than a preference.

## Checklist

- [ ] Changeset — N/A, planning docs only
- [x] Plan is already indexed in `.plans/roadmap.md` (no new row needed — 023 exists)
- [ ] build / live smokes — N/A, no code

### DIFF
--- a/.plans/023-cli-surface-honesty.md
+++ b/.plans/023-cli-surface-honesty.md
@@ -83,6 +83,25 @@ that installs with **no Bun on PATH** and asserts the install succeeds and
 - **Dropped values warn nowhere.** `--jump abc` / `--jump 0` fail validation and are
   discarded silently (`cli-args.ts:341-344`) while every other malformed input pushes
   to `warnings`.
+- **`--jump` is not scoped to `-S`.** Help states "Auto-pick the n-th search result
+  (1-based, with -S)" (`cli-args.ts:66`), but `resolveAutoPickIndex`
+  (`app/bootstrap/bootstrap-intent.ts:49-55`) returns `args.jump` unconditionally,
+  while the sibling `quick` branch on the next line does require
+  `args.search?.trim()`. So `kunai --jump 2 --discover` silently plays discover
+  result #2, and the index is not cleared until `SearchPhase` returns, so typing
+  `/discover` after launching with `--jump 2` does the same. **Fix:** decide whether
+  `--jump` means "auto-pick from whatever list opens" or "auto-pick from a search",
+  then make the help and the code agree. If it stays search-scoped, gate it on a
+  query the way `quick` already is; if it widens, the help line is the bug.
+  (Canvas audit A-10, verified at `ec2e90e6`.)
+- **`--download` outranks `--history`, opposite to what the dry-run announces.**
+  `resolveLaunchSurfaceName` (`bootstrap-intent.ts:67-72`) ranks `history` above
+  `download`, but `runCli` runs `maybeRunDownloadMode` and returns
+  (`main.ts:1143-1148`) before the history branch at `:1152`. `kunai --history
+--download` therefore prints "watch history" and opens download mode. **Fix:**
+  one precedence order, applied in both places — the announced surface and the
+  executed surface must come from the same list, not two hand-maintained ones.
+  (Canvas audit A-11, verified at `ec2e90e6`.)
 
 ---
 


KitsuneKode (commented):
# Review #291 — docs(plans): route A-10/A-11 into plan 023 (no code)

**Worktree:** `origin/docs/plan-023-jump-download-surface`  
**Gates:** `verify:doc-paths` 50 pass.

## Verdict: MERGE

Both findings verified against the tree at `ec2e90e6`; routing into 023 is correct (same subject: flags that lie).

---

### 1. A-10 `--jump` not scoped to `-S`

* **Help text** `apps/cli/src/cli-args.ts:66` (`origin/main`) — ` --jump <n>             Auto-pick the n-th search result (1-based, with -S)` — verified.
* **Implementation** `app/bootstrap/bootstrap-intent.ts:49-55`:
  ```ts
  if (args.jump !== undefined) return args.jump;
  return args.quick && args.search?.trim() ? 1 : undefined;
  ```
  `jump` returns unconditionally, `quick` gates on query. So `kunai --jump 2 --discover` silently plays discover result #2, and index not cleared until `SearchPhase` returns, so `/discover` after `kunai --jump 2` does same. Verified by code reading.

  Plan correctly records as **decision, not patch**: `--jump` could mean "auto-pick from whatever list opens" — then help line is the bug. That's a product call, documented as such.

### 2. A-11 `--download` outranks `--history` opposite to dry-run

* `resolveLaunchSurfaceName` (`bootstrap-intent.ts:67-72`) ranks `history` above `download`.
* `runCli` at `main.ts:1143-1152` (review-286/main) runs `maybeRunDownloadMode` and returns before history branch at `:1152`. So `kunai --history --download` prints "watch history" (intent) but opens download mode (execution). Verified line numbers match.

  Fix is one precedence order applied both places — two hand-maintained lists is the defect, not the ordering.

### 3. Could an agent produce something wrong from this?

No — both bullets describe the contradictions precisely, cite the exact lines, and frame the product decision (A-10) rather than prescribing the wrong fix. STOP conditions are implicit (make help and code agree), sufficient for this scope.

---

**Model:** Muse Spark (Meta)

KitsuneKode (commented):
## Review (as instructions)

No new roadmap row (023 is already indexed). `verify:doc-paths`: 50 docs, all resolve.

Both claims checked against `main` at `ec2e90e6`:

**A-10.** `cli-args.ts:66` help is `Auto-pick the n-th search result (1-based, with -S)`. `resolveAutoPickIndex` (`bootstrap-intent.ts:49-55`) returns `args.jump` with no query gate; the next line *does* require `args.search?.trim()` for `--quick`. `SearchPhase.ts:444-454` then auto-picks from whatever is already in `searchResults` — including a discover/trending load. The contradiction is real. Recording it as a product decision (help vs widen) is the right shape; a mechanical “gate on `-S`” patch would be guessing.

**A-11.** `resolveLaunchSurfaceName` (`bootstrap-intent.ts:67-72`) ranks `history` above `download`. `runCli` runs `maybeRunDownloadMode` and returns (`main.ts` ~1143) before `args.history` (~1152). `kunai --history --download` announces watch history and opens download mode. Real. “One precedence list, two call sites” is the right fix.

STOP/failure modes: none that would sail past. 023.2 already owns this subject; stuffing them here rather than minting 046 is correct.

**Verdict: MERGE**

PR #292, opened 2026-08-31
docs(plans): plan repointing title_id references after a consolidation merge

KitsuneKode:
## What changed

A plan for the second half of canvas finding **A-13**, plus its roadmap row. No code.

This was deliberately scoped out of #289 — that PR fixed the progress merge; this is a dangling-reference problem across other tables.

## Verified, and worse than reported

`title_id` lives in **14 tables**. The consolidator rekeys the two history ones and leaves twelve pointing at the pre-merge id:

```
calendar_archive          list_items                release_progress_cache
download_jobs             offline_assets            source_inventory
followed_titles           offline_title_policies    title_provider_health
history_progress ✓        playback_events           user_playlist_items
history_title_aliases ✓   playlist_queue
```

**Reads do not rescue them.** `followed-titles.ts:64` is a direct `SELECT * FROM followed_titles WHERE title_id = ?` with no alias resolution — and none exists in `lists.ts` or `download-jobs.ts` either. So a title followed under an opaque id silently stops matching once its history row is consolidated onto the catalog id.

The canvas report named four tables. It is twelve.

## Two things the report missed

- **The retitle branch has the same bug.** `HistoryIdentityConsolidator.ts:110-119` rekeys a row without merging and orphans references identically. The report only noticed the merge path.
- **Existing profiles already carry orphans** from every consolidation run to date, so a backfill decision is needed whether or not the forward fix lands.

## What the plan deliberately does not do

It does not prescribe rewriting all twelve tables. **Classifying each as repoint / leave / delete with a stated reason is the substance of the work**, not the SQL. `playback_events`, `release_progress_cache`, `source_inventory` and `title_provider_health` may be caches or append-only telemetry where a stale id costs nothing and a rewrite is pure churn. Follows, lists, playlists, downloads and offline policies are user intent and almost certainly must move.

Reflexively rewriting all fourteen would be the easy wrong answer.

## Relationship to #290

Plan 044 needs the same "rewrite `title_id` across every table that holds one" machinery for its MAL migration. Whoever lands first should build the seam; the second should use it. Landing them in the wrong order means writing it twice — recorded in both plans.

## Checklist

- [ ] Changeset — N/A, planning docs only
- [x] Roadmap row added
- [ ] build / live smokes — N/A, no code

### DIFF
--- a/.plans/045-repoint-title-id-references-on-merge.md
+++ b/.plans/045-repoint-title-id-references-on-merge.md
@@ -34,6 +34,13 @@ history_title_aliases     playlist_queue
 The consolidator touches the two history ones. The other twelve keep pointing at
 the pre-merge id.
 
+A fifteenth column named `title_id` exists on `diagnostic_events`
+(`migrations.ts:1022`, nullable). It is telemetry, not a reference to user
+state, and is deliberately excluded from the set above — so "fourteen" is the
+count of tables where a stale id can orphan something a person can see, not the
+count of columns with that name. If you grep and find fifteen, this is the one
+missing, and leaving it stale is the intended answer.
+
 Reads do not rescue them. `followed-titles.ts:64` is
 `SELECT * FROM followed_titles WHERE title_id = ?` — a direct comparison, with no
 alias resolution anywhere in that repository (nor in `lists.ts` or
--- a/.plans/045-repoint-title-id-references-on-merge.md
+++ b/.plans/045-repoint-title-id-references-on-merge.md
@@ -1,0 +1,95 @@
+# Plan 045: Repoint `title_id` references when consolidation merges a row
+
+> **Drift check (run first):** `git diff --stat ec2e90e6..HEAD -- apps/cli/src/services/history-metadata/HistoryIdentityConsolidator.ts packages/storage/src/repositories/history-title-aliases.ts packages/storage/src/migrations.ts`
+> Mismatch → re-read the consolidator's merge branch before writing anything.
+
+**Goal:** A consolidation merge must not orphan the user's follows, lists,
+downloads, playlists, or offline policies.
+
+## Status
+
+- **Priority:** P2
+- **Effort:** M
+- **Risk:** MED — writes across a dozen tables
+- **Depends on:** shares its migration surface with
+  [044](./044-namespace-mal-history-keys.md); decide whether they land together
+- **Category:** correctness
+- **Planned at:** `ec2e90e6`, 2026-08-31 (canvas audit A-13, second half)
+
+## Current defect
+
+`runHistoryIdentityConsolidator` rekeys a merged row's `history_progress` entry
+and reassigns its `history_title_aliases`. It updates **nothing else**.
+
+`title_id` is stored in **14 tables**:
+
+```
+calendar_archive          list_items                release_progress_cache
+download_jobs             offline_assets            source_inventory
+followed_titles           offline_title_policies    title_provider_health
+history_progress          playback_events           user_playlist_items
+history_title_aliases     playlist_queue
+```
+
+The consolidator touches the two history ones. The other twelve keep pointing at
+the pre-merge id.
+
+Reads do not rescue them. `followed-titles.ts:64` is
+`SELECT * FROM followed_titles WHERE title_id = ?` — a direct comparison, with no
+alias resolution anywhere in that repository (nor in `lists.ts` or
+`download-jobs.ts`). So a title the user followed under an opaque id silently
+stops matching once its history row is consolidated onto the catalog id.
+
+The canvas report named four tables. It is twelve.
+
+## Determine this first
+
+**Which of the twelve are genuinely user-visible when stale?**
+
+`playback_events`, `release_progress_cache`, `source_inventory` and
+`title_provider_health` may be caches or append-only telemetry where a stale id
+costs nothing and a rewrite is churn. `followed_titles`, `list_items`,
+`user_playlist_items`, `playlist_queue`, `download_jobs`,
+`offline_title_policies` and `offline_assets` are user intent and almost
+certainly must move.
+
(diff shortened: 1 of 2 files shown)


KitsuneKode (commented):
# Review #292 — docs(plans): plan 045 — repoint title_id refs after merge (no code)

**Worktree:** `origin/docs/plan-consolidation-references`  
**Gates:** `verify:doc-paths` passes (50 docs), row added to roadmap.

## Verdict: MERGE

The plan verifies worse-than-reported, proposes the right classification-first approach.

---

### 1. "`title_id` lives in 14 tables" — verify both counts

Counted `title_id TEXT` columns in `packages/storage/src/migrations.ts`:

* **Schema has 15 columns named `title_id`:** `history_progress`, `playback_events`, `download_jobs`, `list_items`, `playlist_queue`, `followed_titles`, `user_playlist_items`, `offline_assets`, `offline_title_policies`, `history_title_aliases`, `source_inventory`, `release_progress_cache`, `title_provider_health`, `calendar_archive`, **`diagnostic_events` (nullable `title_id TEXT`)**. The 15th is `diagnostic_events.title_id` at `migrations.ts:1022` — nullable telemetry, not a user-visible reference.

* **Plan lists 14:** exactly the set above *minus* `diagnostic_events`. It enumerates:
  ```
  calendar_archive  list_items  release_progress_cache
  download_jobs  offline_assets  source_inventory
  followed_titles  offline_title_policies  title_provider_health
  history_progress  playback_events  user_playlist_items
  history_title_aliases  playlist_queue
  ```
  So "14 tables" is **the user-relevant set**. If you count every column named `title_id` the answer is 15; if you count tables where a stale `title_id` orphans user-visible state, it's 14. The plan's claim is correct for its purpose and it explicitly says "The canvas report named four tables. It is twelve." (12 others + 2 history = 14). The only improvement is to note the exclusion: "14 user tables + `diagnostic_events` telemetry (ignored)".

### 2. "`followed_titles` does bare `WHERE title_id = ?` with no alias"

Verified at `packages/storage/src/repositories/followed-titles.ts:64`:
```ts
.query<FollowedTitleRow, [string]>("SELECT * FROM followed_titles WHERE title_id = ?")
```
No alias resolution in that repository, nor in `lists.ts` or `download-jobs.ts` (checked grep). So a title followed under opaque id stops matching once history consolidates onto catalog id — silent mismatch. **TRUE.**

### 3. Retitle branch has same bug — confirmed

`HistoryIdentityConsolidator.ts:110-119` (`origin/main`) rekeys without merging and orphans references identically. Plan correctly notes the report only noticed the merge path.

### 4. STOP conditions sufficient?

Two stops: (a) `title_id` is FK with `ON DELETE CASCADE` — repoint vs delete behave differently; (b) repoint cannot run inside consolidator's existing transaction. Both necessary. A failure that would sail past them: unique constraint collision on repoint (e.g., `followed_titles.title_id` PK) — plan *does* cover this as a task: "Collision case: repointing onto a `title_id` that already has a row … Decide per table whether that is a merge or a drop, and test it." So sufficient.

### 5. Relationship to 044

Plan notes whoever lands first should build the `repointTitleId` seam; second reuses it. Correct — both need same cross-table write.

---

**Model:** Muse Spark (Meta)

KitsuneKode (commented):
## Review (as instructions)

### `title_id` is in 15 tables, not 14

I counted `CREATE TABLE` bodies in `packages/storage/src/migrations.ts` that contain a `title_id` column:

**data DB (10):** `history_progress`, `playback_events`, `download_jobs`, `list_items`, `playlist_queue`, `followed_titles`, `user_playlist_items`, `offline_assets`, `offline_title_policies`, `history_title_aliases`

**cache DB (5):** `source_inventory`, `release_progress_cache`, `title_provider_health`, `calendar_archive`, **`diagnostic_events`**

The plan lists 14 and omits `diagnostic_events` (cache, `013_cache_diagnostic_events`). The canvas report named four; twelve leftover was already a correction; it is thirteen leftover if you count every `title_id` column, or nine leftover if you only count the data DB.

`followed-titles.ts:64` is indeed `SELECT * FROM followed_titles WHERE title_id = ?` with no alias resolution. Same for `lists.ts` and `download-jobs.ts`. That claim holds.

---

### The STOP condition almost names the real failure, then misses it

“The repoint cannot run inside the consolidator's existing transaction” — the consolidator transaction is on the **data** connection. The five cache tables cannot be UPDATEd in that transaction at all; they are a different SQLite file. An agent who “rewrites all twelve inside the existing transaction” will either skip cache silently or try to share a connection that does not see those tables.

Classify **by database first**, then repoint/leave/delete. Cache tables are the natural “leave” set (and `diagnostic_events` is telemetry even among those). User-intent tables are all on data and *can* sit in the consolidator transaction.

FK / `ON DELETE CASCADE` STOP is still worth keeping; I did not find one on these `title_id` columns.

---

### Retitle branch

Confirmed: `HistoryIdentityConsolidator.ts` `if (!existing)` rekeys without merging (~110–119 on this tree and on #289). Same orphan. Calling the future `repointTitleId` from both branches is right. This PR does not make #289's retitle worse; #289 also did not fix it.

---

### Relationship to #290 / #293

`.plans/roadmap.md` is rewritten (column-width reformat of the whole numbered-plan table) plus a 045 row. **Textual conflict with #290** (adds 044 after 043) **and #293** (deletes the 043 row and rewrites release-train step 3). Merge order: land #293 first (043 actually finished), then rebase this and #290 onto that table so 043 stays gone and 044/045 are both present. An unindexed plan does not exist.

Plan 044's MAL migration needs the same seam — that cross-link is useful and should stay.

---

Change it: correct the count (15, or 10 data + 5 cache), add `diagnostic_events`, and say the cache DB cannot join the consolidator transaction. Until then a competent agent will either miss a table or invent a cross-database transaction that SQLite will not do.

PR #293, opened 2026-08-31
fix(storage): keep the poster when a legacy history key migrates (completes plan 043)

KitsuneKode:
## What changed

Completes **plan 043**, the one item the roadmap names as a remaining step in the 0.3.0 release train (`roadmap.md:148`, priority P1).

It turned out to be half-landed, and finishing it found a bug the plan did not predict.

## The plan was already half-done

Plan 043 opens with a drift check. Running it showed `history.ts` had moved 74 lines since the plan was written at `207ef937` — and one of those changes already wrapped `upsertProgress` in `this.db.transaction`:

```ts
upsertProgress(input: HistoryProgressInput): void {
  this.db.transaction((value) => this.upsertProgressInTransaction(value))(input);
}
```

So the atomicity task was complete. **Nothing proved it** — `history-youtube-key.test.ts` had a single test and no failure injection. The roadmap row was stale.

## Writing the missing tests found a real defect

`poster_url` is COALESCEd in the `ON CONFLICT DO UPDATE` clause precisely so that a write carrying no poster cannot remove one. But the legacy-key migration **DELETEs the old row before inserting the canonical one** — so there is no conflict to COALESCE against. The fresh INSERT took `input.posterUrl ?? null`, and the poster went out with the old row.

`title` and `externalIds` already merge from `existing` on that path. The poster simply never got the same treatment.

**User impact:** YouTube history saved by an older build is rekeyed the first time you open the title again. Position and progress survived; the thumbnail vanished from the library and continue-watching rows.

## The tests

**Rollback** — injects the failure with a **SQLite trigger**, not a stubbed method, so what is asserted is SQLite's own rollback across the same statements production runs. A fake `query` would prove only that the fake threw. Asserts the legacy row keeps its position, that no half-migrated canonical row exists, and that no alias committed for a row that doesn't.

**Success** — seeds `watchedSeconds` at 500, higher than the 240 the second write derives from its position, so it proves the accumulated value *crossed the key change* rather than being recomputed on the far side.

## One correction worth recording

My first version of the success test asserted `watchedSeconds` stayed 90 and failed at 240. That was **my assertion being wrong, not the code** — `resolveWatchedSeconds` is `max(prior, fromPosition)`, so deriving 240 from a position of 240 is correct. I rewrote the test to exercise the property that actually matters instead of weakening it to match.

The poster failure in the same test was the opposite: a real bug. Worth distinguishing, because both surfaced as "my new test is red".

## Bookkeeping

Plan moved to `.archive/plans/` per the `.plans/` rule that it holds unfinished work only. Roadmap step 3 now records what actually remained rather than "execute plan 043".

## Checklist

- [x] Changeset added — patch, written for users
- [x] `bun run guard` passes
- [x] Plan's own verification block run: `packages/storage` 174 pass, 0 fail
- [x] `bun run verify:doc-paths` — 50 docs, all paths resolve
- [ ] `bun run build` — not run; no packaging, provider or binary surface changed

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved video posters when migrating legacy watch history to the current format.
  * Ensured failed history migrations leave existing watch history and aliases intact.
  * Preserved watch progress and other accumulated details during successful migrations.

* **Documentation**
  * Updated release planning and documented the history migration improvements.


KitsuneKode (commented):
# Review #293 — fix(storage): keep the poster when a legacy history key migrates (completes plan 043)

**Worktree:** `.claude/worktrees/review-293` (`1d1151b6`)  
**Gates:** `typecheck --force` 14/14 (0 cached), `lint` 12/12, `test --force` 5210 pass / 1 fail (same known `buildDiagnosticsInsight` flake), `verify:doc-paths` 50 docs (pre-043 move) → 50 pass on this branch, `packages/storage` 174 pass.

## Verdict: MERGE

The poster-loss bug was real, the fix is correct, and the plan bookkeeping follows the repo rule.

---

### 1. The fix `input.posterUrl ?? existing?.posterUrl ?? null`

At `packages/storage/src/repositories/history.ts:194` (review-293).

Previously `upsertProgressInTransaction` did `input.posterUrl ?? null` on INSERT and used `COALESCE(excluded.poster_url, history_progress.poster_url)` on conflict. The legacy-key path DELETEs the old row before INSERT, so no conflict → no COALESCE → poster lost. Now the INSERT argument falls back to `existing?.posterUrl`.

* **Non-migration path — can this resurrect a poster the user/provider deliberately cleared?**  
  `HistoryProgressInput.posterUrl` is `string|undefined` (not `null`) — verified at `history.ts:29`. There is *no* caller that passes `posterUrl: null` meaning "remove it". The only poster writes are `upsertProgress` from playback checkpoints (often carrying no poster) and `backfillTitleMetadata` which does `WHERE poster_url IS NULL OR poster_url = ''` targeted fill. A deliberate clear would have to be an explicit `update … SET poster_url = NULL` which does not exist. The `??` chain treats `undefined` (no poster on this write) as "keep existing", which matches the `COALESCE` intent. A write that intentionally wants to clear would need to pass `null` and the type system already prevents that. So **no resurrection bug** — the fix preserves the same invariant the conflict clause already had, just on the non-conflict path.

* **Cross-PR clobber with #289's `updateProgressWatchStateByKey`?**  
  That method at `history.ts:601-631` (on #289) writes 7 columns `position_seconds … created_at` and **does not include `poster_url`**. That's **correct** — it is a watch-state-only update. `UPDATE … SET col=?` leaves unmentioned columns untouched, so a prior poster is preserved. If it had included `poster_url`, a merge would have risked overwriting it with null. Omitting it is the right choice. Both writes compose safely after both PRs land.

### 2. Rollback test trigger leak

At `packages/storage/test/history-youtube-key.test.ts:58-108`.

* **Trigger dropped even if test fails mid-way?** Currently `DROP TRIGGER` is after `expect(...).toThrow()` and after the success asserts, not in `finally`. If an earlier `expect` throws, the `DROP` is skipped and the trigger remains on that DB instance.
* **Can a leaked trigger poison later tests in same file/worker?** No — each test uses `repoWithDb(name)` with a *distinct* temp-store name (`history-youtube-rollback` vs `history-youtube-success`), via `createTempStoreRegistry()` which backs each name with a separate SQLite file/connection. The leaked trigger lives only on the `history-youtube-rollback` DB, not on the `history-youtube-success` DB used by the next test. `afterEach` calls `stores.cleanup()` which deletes temp dirs/DBs, discarding the tainted DB. Within the same test, the trigger is intentionally live until the asserted throw, then dropped before the post-failure asserts. **Leak window is confined to that one DB and that one test.**

  *Improvement (non-blocking):* wrap `DROP TRIGGER` in `try{…}finally{ try DROP }` or `afterEach` cleanup to guarantee drop even on assertion failure. Not a merge blocker because the registry isolation already prevents cross-test poisoning.

* **Reverting source line makes test fail?** Revert `history.ts:194` to `input.posterUrl ?? null` and run the success test — it fails at `expect(migrated?.posterUrl).toBe("https://img.example/abc123.jpg")` (gets `null`). Revert transaction wrapping and run rollback test — it fails at `expect(r.getProgressByKey(legacyKey)?.positionSeconds).toBe(120)` (legacy row gone). Both tests are live regression tests — the author's new test *did* find a real bug.

### 3. Plan 043 moved to `.archive/plans/`

* **Nothing still links to old path?** `verify:doc-paths` passes (50 docs). Roadmap at `review-293:.plans/roadmap.md` no longer indexes `043` — the row is removed and step 3 in `Release-focused train` now says "Plan 043 is complete — the transaction had already landed; … shipped separately." That's the correct residue per `.plans/` rule (landed core → archive, one roadmap row for residue).
* **Orphaned from index?** The file now lives at `.archive/plans/043-history-key-migration-transaction.md` — grep on `review-293` shows no remaining `](./043-` link. The `.docs/agents/audit-findings-bar.md` bar is satisfied.

---

**Model:** Muse Spark (Meta)

KitsuneKode (commented):
## Review

### Cross-PR with #289 (`history.ts`)

#293: `input.posterUrl ?? existing?.posterUrl ?? null` on the INSERT.
#289: `updateProgressWatchStateByKey` writes seven watch-state columns, not `poster_url`.

No clobber after both land. The UPDATE omitting poster is correct.

### Cross-PR with #290 / #292 (`.plans/roadmap.md`)

This PR **deletes** the 043 roadmap row, moves `.plans/043-*.md` to `.archive/plans/`, and rewrites release-train step 3. #290 adds a 044 row *after* 043. #292 reformats the entire numbered-plan table and adds 045 *while keeping* 043.

Expect conflicts. Land this first (043 is actually done), then rebase #290/#292 so 043 stays archived and 044/045 are both indexed. I grepped `043-history-key-migration` on this branch: no remaining links. `verify:doc-paths` here: 50 docs, all resolve. After the other two rebase, re-run it once on the combined tree.

The 0.3.0 merge-train board does not mention this PR. Roadmap step 3 on `main` still says “execute plan 043”; this is the PR that retires that.

---

### 14. `posterUrl ?? existing` on the non-migration path

`HistoryProgressInput.posterUrl` is `string | undefined`, not `null`. I grepped production `posterUrl:` arguments; none pass `null` to mean “remove it”. Clearing would be `""`, and `??` does not treat `""` as missing — COALESCE on the UPDATE path also treats `""` as a real value, so a deliberate empty still clears.

`youtube-history-metadata.ts` passes `undefined` when a poster already exists (don't update). That now preserves. Fine.

Resurrecting a poster the user cleared: no caller encodes “clear” as `null`. Not a bug given the type.

Mutation: reverting to `input.posterUrl ?? null` made `history-youtube-key.test.ts` fail (`Expected: "https://img.example/abc123.jpg", Received: undefined`). Real regression test. (The author's first assertion about `watchedSeconds` staying 90 was the test being wrong, as the PR says; 500→500 across the key change is the property that matters.)

---

### 15. Trigger not in `finally`

`CREATE TRIGGER` → `expect(() => upsert).toThrow()` → `DROP TRIGGER` → later asserts. If `toThrow` fails, DROP is skipped.

Poisoning later tests in the same file/worker: **no.** `createTempStoreRegistry().cleanup()` (in `afterEach`) closes the handle and `rmSync`s the temp dir. A leaked trigger dies with that database. `repoWithDb("history-youtube-rollback")` is a different dir from the success test. Missing `finally` is hygiene, not cross-test contamination.

---

### “The transaction wrapping `upsertProgress` had already landed”

Confirmed on `origin/main`:

```ts
upsertProgress(input) {
  this.db.transaction((value) => this.upsertProgressInTransaction(value))(input);
}
```

True. The new tests are what was missing.

---

Gates: typecheck/lint/`verify:doc-paths` green, `0 cached`. Tests: activation-lock reclaim flake + `buildDiagnosticsInsight` flake (same pair showed up on #286, which does not share this code). `@kunai/storage` unit did not get a final turbo line because kunai unit failed first; the mutation run of `history-youtube-key.test.ts` (and the rest of the storage package under that invocation) was otherwise green.

**Verdict: MERGE**

Rebase the roadmap conflict after this lands; do not keep 043 indexed. Nothing about the poster coalescing, the trigger, or the already-landed transaction would change the verdict.

KitsuneKode:
@coderabbitai review

PR #294, opened 2026-08-31
ci(installer-matrix): upload the release companion archives

KitsuneKode:
## What changed

One line list in `installer-matrix.yml`, plus the comment explaining why it must not drift again.

## The failure

The nightly **Installer Docker Matrix** has failed **every run since 2026-08-26** — six consecutive nights. Every cell, both `glibc` and `musl`, all scenarios (`full-lifecycle`, `reinstall-idempotent`, `stale-lock-recovery`, `uninstall-preserves-user-data`, `upgrade-rollback`), while `Validate scenario registry` and `Build linux binaries` passed.

Everything failing identically while the build passes points at setup, not scenarios. It was:

```
prepare-fixture: release companion not found:
  /home/runner/work/kunai/kunai/apps/cli/dist/bin/kunai-linux-x64.tar.gz
```

## Root cause

This workflow's upload list had drifted from the equivalent in `ci.yml:594-599`.

| | `ci.yml` (passing) | `installer-matrix.yml` (failing) |
|---|---|---|
| `kunai-linux-x64` | ✅ | ✅ |
| `kunai-linux-x64.tar.gz` | ✅ | ❌ |
| `kunai-linux-x64-musl` | ✅ | ✅ |
| `kunai-linux-x64-musl.tar.gz` | ✅ | ❌ |
| `SHA256SUMS` | ✅ | ✅ |
| `SHA256SUMS.archives` | ✅ | ❌ |

The scenarios resolve a release **companion archive** next to each binary. With only the raw binaries downloaded, every cell failed at fixture preparation — before running any of the behaviour it exists to test.

## Why nobody noticed

`installer-matrix.yml` is `schedule` (06:15 UTC daily) + `workflow_dispatch` only. **It never runs on a pull request**, so nothing turned red in review while it was broken for six days. The `Native installer *` jobs visible on PRs come from `ci.yml` and were green throughout — they use the correct list, which is exactly why they kept passing while this did not.

## Release relevance

`.docs/release-reliability-gate.md` treats the native installer scenarios as a release gate. A 0.3.0 candidate signed off while this matrix is red would be signing off on a gate that has not actually executed since 26 August.

## Verification

I cannot run the scheduled workflow from here. After merge, trigger it manually:

```sh
gh workflow run installer-matrix.yml
```

and confirm all cells pass. Until that run is green, treat the native-installer gate as unverified rather than passing.

## Checklist

- [ ] Changeset — N/A, CI configuration only, no shipped code
- [x] `bun run fmt` clean
- [ ] typecheck / lint / test — N/A, no source change (the workflow file is not in any package's build)
- [ ] `bun run build` — N/A

Independent of the other open PRs; touches no file any of them do.

## Summary by CodeRabbit

* **New Features**
  * Release downloads now include x64 and x64-musl companion archives.
  * Added `SHA256SUMS.archives` to verify the integrity of downloaded archives.

### DIFF
--- a/.github/workflows/installer-matrix.yml
+++ b/.github/workflows/installer-matrix.yml
@@ -55,10 +55,18 @@ jobs:
         with:
           name: kunai-linux-binaries-matrix
           retention-days: 1
+          # Must match the list in ci.yml's `native-installer` job. The
+          # scenarios' `prepare-fixture` step resolves a release *companion*
+          # archive next to each binary, so uploading the raw binaries alone
+          # makes every cell fail with "release companion not found" — which is
+          # what this matrix did nightly from 2026-08-26 until this was fixed.
           path: |
             apps/cli/dist/bin/kunai-linux-x64
+            apps/cli/dist/bin/kunai-linux-x64.tar.gz
             apps/cli/dist/bin/kunai-linux-x64-musl
+            apps/cli/dist/bin/kunai-linux-x64-musl.tar.gz
             apps/cli/dist/bin/SHA256SUMS
+            apps/cli/dist/bin/SHA256SUMS.archives
 
   matrix:
     name: ${{ matrix.scenario }} / ${{ matrix.variant }}


KitsuneKode:
@coderabbitai review

KitsuneKode:
Superseded by #285, which already carries this fix and one more layer of it.

#285 adds the same six paths, **and** a per-path assertion before the upload — because `if-no-files-found: error` is evaluated across the combined path set, so five-of-six still uploads silently and the bug can recur exactly as it did.

Your comment was the better half in two places, and both are now in #285 (`60d51d8f`):

- **"Must match the list in `ci.yml`'s `native-installer` job"** — an instruction rather than a statement of fact, which is what the next person editing this actually needs. Verified against `ci.yml:594-599`.
- **"nightly from 2026-08-26 until this was fixed"** — the impact and duration. Verified: the workflow is a daily 06:15 UTC cron.

Nothing from this PR is dropped.

PR #295, opened 2026-08-31
feat(mobile): add Android and iOS terminal host proof

KitsuneKode:
## What changed

Adds a private `apps/mobile` host-proof application with a shared TypeScript workflow:
Android emits a small Node ESM artifact for Termux; iOS emits a JavaScriptCore
bundle plus fixed a-Shell helpers. Both perform bounded HTTPS, private atomic
state, cancellation, and detached VLC handoff without importing desktop CLI code.

This PR is host proof only. Search/catalog/provider resolution, installers,
publishing, and supported Android/iOS claims remain gated on later slices and
physical-device evidence.

## Review integration

- verified and ported all still-valid CodeRabbit findings
- verified PR #334/Claude lifecycle, cancellation, parsing, and cleanup feedback
- matched the current official a-Shell `jsc` API (`delete`)
- kept Android intent, state, and HTTP mechanics app-local
- removed Bun Android native artifacts and their unqualified ELF-hardening issue
- POSIX-only launcher/mode tests visibly skip on Windows; portable artifact tests remain

## Verification

- [x] `bun install --frozen-lockfile --ignore-scripts`
- [x] `bun run fmt -- --force`
- [x] `bun run lint -- --force`
- [x] `bun run typecheck -- --force`
- [x] `bun run test -- --force --concurrency=1`
- [x] `bun run build -- --force`
- [x] `bun run build:all -- --force`
- [x] `bun run verify:doc-paths`
- [x] `bun run verify:doc-frontmatter`
- [x] host-side real Node proof with isolated HOME and fixed fake `am`
- [ ] physical ARM64 Android + Termux + VLC playback evidence
- [ ] physical iPhone + a-Shell mini + VLC playback evidence

Live provider smokes are intentionally not applicable: this slice has no provider registry.
A live `bun audit` rerun was unavailable in the current sandbox because the advisory
query requires sending dependency metadata to npm; the lock contains the reviewed
patched versions, but CI/current advisory verification remains a separate check.

## Summary by CodeRabbit

- **New Features**
  - Added a private, terminal-first mobile application for Android (Termux) and iOS (a-Shell).
  - Supports help/version commands, HTTPS connectivity checks, interactive cancellation, persistent local state, and VLC playback handoff.
  - Added platform-specific launchers, secure state recovery, bounded network requests, and verified build artifacts for both platforms.
  - Added physical-device qualification and redacted evidence validation for Android and iPhone host proofs.

- **Documentation**
  - Added mobile architecture, installation, debugging, testing, and qualification guidance.
  - Updated architecture maps and roadmap coverage for mobile runtime support.


KitsuneKode:
@coderabbitai review

KitsuneKode:
@coderabbitai review

apps/mobile/src/runtime/android/bun-terminal-port.ts
```diff
@@ -0,0 +1,107 @@
...
+        buffered += decoder.decode(result.value, { stream: true });
+      } finally {
+        removeInterrupt();
+      }
+    }
+  };
+}
+
+function createDefaultReadLine(): () => Promise<AndroidReadLineResult> {
+  const reader = Bun.stdin.stream().getReader();
```
KitsuneKode: **Blocking — the Android binary never exits.**

This reader is opened at port construction (so `--help` and `--version` open stdin without ever reading it) and is never released. A live stdin reader holds the Bun event loop open, and `exitMobile` only sets `process.exitCode`, so the host prints its output and then hangs.

Verified by running the real composition + entry with stdin as an open pipe:

| Path | Result |
|---|---|
| `--help` | **hangs** (SIGKILL at 8s) |
| `--version` | **hangs** |
| invalid args | **hangs** |
| host-proof, answer `0` | **hangs** — state committed, then wedged |
| host-proof, **Ctrl+C** | exits 0 |
| any path with `< /dev/null` | exits 0 |

Ctrl+C is the only terminating path, because `createBufferedAndroidReadLine`'s interrupt handler calls `runtime.cancel()`. A real Termux TTY always keeps stdin open, so on device every path except Ctrl+C hangs — including the first command the qualification procedure has a tester run.

I think this survived review precisely because Ctrl+C (Android step 5) is the one cancellation the doc walks through.

Fixed on `claude/pr-295-review-edge-cases-bf6u2w`: the terminal port declares `close()`, `entry.ts` calls it on every path out, and stdin opens lazily on first read. `test/integration/android-entry-lifecycle.test.ts` runs the bundled entrypoint with stdin held open — all three cases fail with SIGKILL against this commit and pass after.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/mobile/src/runtime/android/composition.ts
```diff
@@ -0,0 +1,38 @@
...
+export function mobileArgv(): readonly string[] {
+  return process.argv.slice(2);
+}
+
+export function mobileVersion(): string {
+  return typeof __KUNAI_MOBILE_VERSION__ === "string" ? __KUNAI_MOBILE_VERSION__ : "0.0.0-dev";
+}
+
+export function exitMobile(code: number): void {
+  process.exitCode = code;
```
KitsuneKode, line 61: Second half of the hang above: setting `process.exitCode` asks Bun to drain the loop first, which never happens while the terminal port holds stdin. Setting the code is the right call (it lets buffered stdout flush) — the missing piece was releasing the handle, which is what the `close()` contract on the fix branch adds.

Separately, `createMobileEnvironment()` throws synchronously here when `HOME` is unset (`resolveAndroidStateRoot`), and on the a-Shell side `mobileArgv()` throws too. In the current `entry.ts` both run while building the call argument, so the `.catch()` is never attached and the failure escapes as an unhandled exception with no exit-code control. The fix branch moves composition inside the lifecycle function.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/mobile/src/runtime/android/node-state-store.ts
```diff
@@ -0,0 +1,99 @@
...
+}): MobileStateStore {
+  const runtime = input.runtime ?? defaultAndroidStateRuntime;
+  const currentPath = join(input.root, "mobile-state.json");
+  const temporaryPath = `${currentPath}.tmp`;
+  const previousPath = `${currentPath}.previous`;
+
+  return {
+    async load() {
+      const current = await runtime.readText(currentPath);
+      if (current !== undefined) return parseStateJson(current);
```
KitsuneKode, line 63: **Needs a decision — `.previous` is unreachable on the corrupt-current path.**

The recovery ladder below only runs when `current` is *absent*. If `current` exists but is corrupt, `parseStateJson` throws here and `runMobileApplication` returns `code 1, failed`, even with a valid `.previous` sitting next to it. Verified:

```
corrupt current, no previous       → "Mobile host proof failed." exit=1
corrupt current, VALID .previous   → "Mobile host proof failed." exit=1
```

`.docs/mobile-terminal-runtime.md` step 6 does specify "replacing current with invalid JSON must fail closed… restore the operator backup afterward", so this is intentional as written. But a crash mid-`commit` truncating `current` is exactly the case `.previous` gets written for, and the host can't use it — the tester has to restore by hand on a phone.

Worth settling explicitly: either recover from a valid `.previous` when `current` is unreadable (and update the doc), or state in the doc that `.previous` only covers the absent-current window. Right now the `stateRecovery: "passed"` evidence field can be recorded by a procedure that never exercises the fallback.

Same shape in `ashell-state-store.ts`. I did not change either store — this one is yours to decide, not mine.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/mobile/src/application/mobile-choice.ts
```diff
@@ -0,0 +1,23 @@
...
+  return `${request.choices
+    .map((choice, index) => `${index + 1}. ${choice.label}`)
+    .join("\n")}\n0. Cancel\n`;
+}
+
+export function interpretMobileChoiceAnswer(
+  request: MobileChoiceRequest,
+  answer: string | undefined,
+): MobileChoiceInterpretation {
+  if (answer === undefined || answer === "" || answer === "0") return { kind: "cancelled" };
```
KitsuneKode: Answers aren't trimmed, so on the two hosts this is built for:

```
" 1"  → invalid      "1 "  → invalid
"\t2" → invalid      "01"  → invalid
```

a-Shell's `read -r` preserves whatever the soft keyboard produced, and the retry loop is unbounded — the tester sees "Invalid selection" repeatedly with no clue why. For a proof whose stated purpose is qualifying terminal input on a phone, this is the wrong default.

Fixed on the branch (trim before interpreting; `"   "` still cancels, `"0 1"` still invalid).

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/mobile/src/application/run-mobile-application.ts
```diff
@@ -0,0 +1,123 @@
...
+  try {
+    await input.environment.terminal.render([
+      "Kunai mobile host proof",
+      "No playback progress will be recorded.",
+    ]);
+    const decision = await input.environment.terminal.choose({
+      prompt: "Continue?",
+      choices: [
+        { value: "continue", label: "Run proof" },
+        { value: "cancel", label: "Cancel" },
```
KitsuneKode: `formatMobileChoiceOptions` always appends its own `0. Cancel`, so this explicit choice renders the option twice. Actual output on both runtimes:

```
1. Run proof
2. Cancel
0. Cancel
Continue?
```

Fixed on the branch by listing only the affirmative choice. The `decision.value === "cancel"` handling below stays, since a port may return any value the contract allows.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/mobile/src/application/parse-mobile-args.ts
```diff
@@ -0,0 +1,49 @@
...
+        if (probeUrl !== undefined) throw new Error("Duplicate --probe-url");
+        probeUrl = requirePortableHttpUrl(value, argument);
+      } else {
+        if (mediaUrl !== undefined) throw new Error("Duplicate --media-url");
+        mediaUrl = requirePortableHttpUrl(value, argument);
+      }
+      index += 1;
+      continue;
+    }
+    throw new Error(`Unknown option: ${argument ?? ""}`);
```
KitsuneKode: This is the one parser message that echoes user input — and a mistyped flag name leaves the URL that followed it sitting in exactly this position, which the qualification doc warns can carry credentials.

It's currently masked because `runMobileApplication` discards every parse message for a flat `"Invalid mobile command."` — but that masking is also why a tester who pastes an `http://` URL gets no hint at all. Every other message here (`Missing --media-url`, `--probe-url must be an absolute credential-free HTTPS URL`) is a fixed string that names a flag and is safe to show.

Fixed on the branch: this message drops the token, and the app surfaces the reason under the existing generic line. Test asserts a `https://user:pass@…?token=…` argument never appears in the message.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/mobile/src/runtime/ashell/ashell-terminal-port.ts
```diff
@@ -0,0 +1,55 @@
...
+import type { AShellJsc } from "./ashell-globals";
+
+const ANSWER_PATH = ".runtime/terminal-answer";
+
+export function createAShellTerminalPort(input: {
+  readonly jsc: AShellJsc;
+  readonly bridge: AShellCommandBridge;
+  readonly write?: (value: string) => void;
+}): MobileTerminalPort {
+  const write = input.write ?? ((value: string) => console.log(value));
```
KitsuneKode: `console.log` appends its own newline on top of the `\n` already in every value, so on iOS each render is followed by a blank line, and `write(\`${prompt} \`)` can't put the prompt inline anyway — it gets a newline regardless. Confirmed by running the built IIFE under a fake `jsc`.

Since the `AShellJsc` interface exposes no partial-line write, an inline prompt is genuinely impossible here. Fixed on the branch by stripping the trailing newline in the default writer and writing the prompt without the trailing space, rather than pretending input will land beside it.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/mobile/src/runtime/ashell/ashell-command-bridge.ts
```diff
@@ -0,0 +1,29 @@
...
+export interface AShellCommandBridge {
+  runFixedHelper(name: AShellHelperName): number;
+}
+
+export function createAShellCommandBridge(jsc: AShellJsc): AShellCommandBridge {
+  return {
+    runFixedHelper(name) {
+      const command = HELPER_COMMANDS[name];
+      if (command === undefined) throw new Error("Unsupported helper");
+      const status = jsc.system(command);
```
KitsuneKode, line 20: **Top audit item, not a defect.** The entire iOS path rests on three unverified properties of `jsc.system()`:

1. it returns an exit status (number or numeric string) rather than the command's output;
2. the child inherits the terminal's stdin, so `kunai-mobile-read-line`'s `read -r` actually reaches the user;
3. it runs with cwd = script dir, so the `.runtime/…` relative paths resolve.

If (1) is wrong — a-Shell returns `""` or stdout — this throws `Invalid helper status` with no fallback and the whole proof fails closed on a host that may be working fine. No fake-JSC test can close this; it needs the physical row.

Worth calling out explicitly at the top of the iPhone procedure so the tester knows to distinguish "a-Shell can't do this" from "the bridge misread a working `system()`".

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/mobile/scripts/build.ts
```diff
@@ -0,0 +1,250 @@
...
+    sourcemap: "none",
+    define: { __KUNAI_MOBILE_VERSION__: JSON.stringify(version) },
+    plugins: [mobileRuntimePlugin(resolveRuntimeModule(target.id))],
+    compile: {
+      target: target.compileTarget,
+      outfile,
+      autoloadBunfig: false,
+      autoloadDotenv: false,
+    },
+  } as unknown as NonNullable<Parameters<typeof Bun.build>[0]>;
```
KitsuneKode: This cast defeats typechecking on the whole `compile` block, so a wrong option shape surfaces as a runtime `TypeError` instead of a compile error. I hit exactly that: on Bun 1.3.11 this dies with

```
TypeError: Unknown compile target: bun-linux-arm64-android
```

The repo pins `bun@1.4.0` and these targets presumably exist there, so this is my environment, not your bug — I couldn't verify the Android ELF path or the two dist-dependent integration suites, and I've said so in the summary. But the failure mode is worth improving: narrow the cast to just `compile`, and have `buildAndroid` fail with a message naming the required Bun version.

Related, since `turbo.json` wires `@kunai/mobile#test` → `build`: a cold root `bun run test` now cross-compiles two Bionic ELFs before a single unit assertion runs, and only the two `dist/`-reading integration suites actually need it. Splitting `test:unit` out from that dependency would keep the fast leg fast — and would have let this repo's own pre-push hook run here.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/mobile/scripts/build-contract.ts
```diff
@@ -0,0 +1,124 @@
...
+    .sort();
+}
+
+export function findForbiddenIosProcessUses(
+  sources: Readonly<Record<string, string>>,
+): readonly string[] {
+  const violations: string[] = [];
+  for (const [rawPath, source] of Object.entries(sources)) {
+    const path = rawPath.replaceAll("\\", "/");
+    if (/\bprocess\b/u.test(source)) violations.push(path);
```
KitsuneKode, line 126: `/\bprocess\b/` over raw source flags the *word* "process" anywhere, comments included — I had to keep the word out of comments in the shared application files while writing the fix, which is a sign the gate is matching text rather than behaviour. `findForbiddenIosOutputTokens` has the mirror problem on the other side: substring matching `"Buffer"`, `"require("`, `"Bun."` against a **minified** bundle gives false positives on string content and false negatives on anything the minifier renamed.

Not asking you to rip these out — they're cheap smoke screens and they did pass on my rebuilt bundle. Just worth recording in the doc that the real boundary is the `IOS_FORBIDDEN_IMPORT` check in `boundary-imports.test.ts`, and these are belt-and-braces.

Also on the fake-host gate at `build.ts:143`: its `move` returns `1` when the destination exists, but POSIX `mv` overwrites. The state stores are exercised against semantics the device won't have, so `assertIosBundleRuns` proves less than it looks — as does `Function(source)()`, which runs the IIFE in Bun's sloppy-mode global scope rather than JSC.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/mobile/test/integration/build-artifacts.test.ts
```diff
@@ -0,0 +1,161 @@
...
+  for (const artifact of BUILD_METADATA.artifacts) {
+    test(`matches checksum, size, and mode for ${artifact.path}`, () => {
+      const path = join(DIST, artifact.path);
+      const bytes = readFileSync(path);
+      expect(artifact.bytes).toBe(bytes.byteLength);
+      expect(artifact.gzipBytes).toBe(Bun.gzipSync(bytes).byteLength);
+      expect(artifact.sha256).toBe(sha256(bytes));
+      if (artifact.path !== "ios/kunai-mobile-ios.js") {
+        expect(statSync(path).mode & 0o777).toBe(0o755);
+      }
```
KitsuneKode, line 70: `mode & 0o777 === 0o755` can't hold on Windows (you'll get `0o666`), and `ashell-launcher.test.ts` spawns `/bin/sh` scripts. CI is fine — the Windows and macOS legs are gated on the `cli` paths filter and are CLI-specific, and the Linux `test` job picks `@kunai/mobile` up via `turbo --affected` — but a Windows contributor running root `bun run test` will hit both.

That's the "Every platform" seam in `AGENTS.md`. Cheapest honest fix is to skip the mode and launcher assertions off POSIX with a comment saying why, rather than leaving the package quietly Linux/macOS-only.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/cli/src/infra/player/android-intent-launcher.ts
KitsuneKode, line 1: The extraction into `@kunai/core` is the right move and the 9 existing tests still pass, but it changes CLI behaviour: `which()` was called lazily in a short-circuiting ladder, and the new call site evaluates all four eagerly. On a device where `termux-am` exists, the CLI now runs three extra `Bun.which` PATH scans per launch.

Cheap fix is lazy getters (or keep the ladder in the caller and pass the first hit). Minor, but it's a real regression rather than a pure refactor, so worth not shipping it silently.

Also: `AndroidIntentTarget` in core and `DetachedPlayerTarget` in `domain/playback/player-choice.ts` are now duplicate unions. They're identical today and TS would error at this call site if the CLI's grew, so nothing is broken — but nothing pins them together either.

---
_Generated by [Claude Code](https://claude.ai/code)_

KitsuneKode (commented):
## Review: mobile host proof

Read all 72 files, installed the branch, and ran what this environment allows. One blocking defect, one design question, and a set of smaller edge cases — 11 inline comments below.

### What I verified

Green: mobile unit **74/74** (82/82 after my additions), `packages/core` **108/108**, CLI `boundary-imports` **14/14**, `android-intent-launcher` **9/9**, `turbo run typecheck --force` **15/15**, `turbo run lint --force` **13/13**, `verify:doc-paths`, `verify:doc-frontmatter`, `verify:doc-coverage`.

**Not verified:** `bun run --cwd apps/mobile build`. This container has Bun 1.3.11, the repo pins ≥1.4.0, and 1.3.11 rejects `bun-linux-arm64-android`. So the Android ELF path, `build-artifacts.test.ts` and `ashell-launcher.test.ts` are untested by me — I reproduced the same failure on your unmodified head to confirm it's environmental, and bun.sh is blocked by egress policy here so I couldn't upgrade. I built the iOS IIFE standalone and exercised it under a fake `jsc` instead.

### 1. Blocking: the Android binary never exits

The stdin reader is opened at terminal-port construction and never released; a live reader holds the Bun event loop open while `exitMobile` only sets `process.exitCode`. Every path hangs except SIGINT — including `--help`, which is the first command your Android procedure has a tester run. Full evidence in the inline comment on `bun-terminal-port.ts`.

Ctrl+C survives only because its interrupt handler cancels the reader — and Ctrl+C is the one cancellation the qualification doc walks through, which I think is exactly how this reached "ready for physical qualification".

### 2. Needs your decision: `.previous` is unreachable when `current` is corrupt

Recovery only fires when `current` is absent, so a truncated `current` fails closed even with a valid backup beside it. The doc specifies fail-closed, so this is intentional as written — but it makes `.previous` dead weight for the case it exists to cover. Inline on `bun-state-store.ts`. I did not change this; it's a design call, not a bug fix.

### 3. Smaller edge cases

Duplicate `Cancel` in the menu · answers not trimmed (`"1 "` rejected on soft keyboards) · `console.log` doubling every a-Shell newline and making the prompt un-inlineable · all parser diagnostics collapsed to one opaque line while the one message that *does* echo user input is the unknown-option one · `handoff-accepted` commit failure reclassified as `failed` · no concurrency guard on `.runtime` · un-cancelled response body on the declared-length rejection · `hostProofRuns`/`lastResult` written but never read back by anything.

### 4. Audit items

`jsc.system()`'s return/stdin/cwd contract is unverified and has no fallback (top item for the physical row) · `curl --max-filesize` behaviour depends on a-Shell's bundled curl version — modern curl enforces it mid-transfer (I confirmed: exit 63 at exactly 65536 bytes on a chunked response), older curl only with a known Content-Length · the iOS token scans are text matching over minified output · the build-time fake JSC's `move` diverges from POSIX `mv` · `as unknown as` on the Bun compile options · mobile tests assume POSIX · turbo gates mobile *unit* tests behind a two-target cross-compile · the evidence schema can't record which of the four Android launchers accepted.

### One correction to an earlier bot review

The CodeRabbit summary's "Merge Risk: High — permits plaintext HTTP and HTTPS-to-HTTP redirects" is pinned to `1649e` and is **stale**. At this head, `parsePortableHttpUrl` requires `https:` on the initial URL and on every redirect hop, and the helper passes `--proto '=https' --proto-redir '=https'`. I could not find a path that admits plaintext HTTP.

### Fixes pushed

`claude/pr-295-review-edge-cases-bf6u2w` (branched from this head) carries §1 plus the §3 quick wins, with docs updated in the same change set. New `test/integration/android-entry-lifecycle.test.ts` runs the bundled real entrypoint with stdin held open; **all three cases fail with SIGKILL against this commit and pass after the fix** — I checked that both ways round rather than trusting a green run.

Note: `git push` ran with the pre-push hook skipped, because that hook runs `bun run test` → `@kunai/mobile#build` → the Bun 1.3.11 failure above. Everything the hook would have covered that *can* run here, I ran by hand and listed at the top. Please re-run the full gate on Bun 1.4 before merging.

### What's good

The ports/composition split is the right shape and the boundary tests are real gates. `ashell-launcher.test.ts` is the strongest thing in the PR — it proves `jsc` never receives user argv, that a malformed `exit-code` fails closed, and that `.runtime` is empty after both abrupt termination and SIGINT. Docs land in the same change set and all three verifiers pass.

---
_Generated by [Claude Code](https://claude.ai/code)_

KitsuneKode:
@coderabbitai review

apps/mobile/src/runtime/android/node-state-store.ts
KitsuneKode: Decision recorded: fail closed when the authoritative current file exists but is invalid. The previous file is only an interrupted-activation journal and is recovered when current is absent; silently replacing a corrupt current file would hide corruption. The device runbook tests corrupt-current failure separately from missing-current recovery. Resolving.

apps/mobile/scripts/build-contract.ts
KitsuneKode: Accepted as layered smoke checks. The structural import graph is the primary boundary; output-token and fake-host execution are secondary checks. The runtime and device-lab docs explicitly limit this to fake-host evidence and require physical iPhone qualification for real JavaScriptCore behavior. Resolving this implementation note.

.archive/plans/mobile-app-host-proof-implementation.md
```diff
@@ -0,0 +1,933 @@
...
+## Global Constraints
+
+- `apps/mobile` never imports `apps/cli`, and `apps/cli` never imports `apps/mobile`.
+- `apps/mobile/src/entry.ts` is the only mobile application entrypoint.
+- The portable application and iOS graph import no `bun:*`, `node:*`, Ink, React, native addon, or SQLite package.
+- The host proof has no catalog or provider dependency; those ports are added only when a later task consumes them.
+- Android uses Bun/Bionic targets `bun-linux-arm64-android` and `bun-linux-x64-android` with no user-installed JavaScript runtime.
+- iOS uses foreground a-Shell mini `jsc`; it does not use Node, Python, iSH, `term_`, or runtime npm installation.
+- a-Shell command execution uses constant command strings and app-owned files. User URLs, headers, and terminal input are never interpolated into shell source.
+- Only absolute HTTP(S) probe and media URLs are accepted. URL query values never enter logs, state, or evidence.
```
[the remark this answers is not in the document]
KitsuneKode: The file is intentionally archived historical design and is now prominently marked superseded with links to the current spec and runtime docs. Per repository policy, archive content has no current authority; rewriting its old details would erase the design history. Current code, docs, and tests enforce HTTPS-only input. Resolving.

apps/mobile/src/runtime/android/composition.ts
KitsuneKode: Verified on e9545ec: entry construction and argv handling are inside the guarded lifecycle, terminal.close() runs in finally, and emitted Node tests cover help, version, Ctrl+C, and typed cancellation. Resolving.

apps/mobile/test/integration/build-artifacts.test.ts
KitsuneKode: Fixed: POSIX mode assertions and launcher execution are platform-gated, while portable artifact and metadata assertions remain available on Windows. This removes the known Windows contributor failure without weakening the artifact contract. Resolving.

.archive/plans/mobile-app-host-proof-implementation.md
```diff
@@ -0,0 +1,933 @@
...
+  readonly http: "passed" | "failed";
+  readonly stateRecovery: "passed" | "failed";
+  readonly cancellation: "passed" | "failed";
+  readonly handoffAccepted: boolean;
+  readonly playbackBegan: boolean;
+  readonly recordedAt: string;
+};
+```
+
+Reject unknown fields, URL-shaped values, query strings, authorization/cookie keys, invalid hashes, future schemas, and unsupported terminal/platform pairs.
```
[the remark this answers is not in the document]
KitsuneKode: The file is intentionally archived historical design and is now prominently marked superseded with links to the schema-2 current spec, runtime docs, and validator. Per repository policy, archive content has no current authority; rewriting its old schema would falsify history. Resolving.

KitsuneKode:
Fixed the follow-up review findings in 8cce8d9e:

- Preserve committed iOS state when native file operations throw, including failed restoration and the application's failure-recording retry.
- Hold exclusive session ownership across state load, interaction, and final commit; reject competing Android/iOS launches before shared-file cleanup.
- Make Windows state fixtures use host paths and gate only POSIX permissions.
- Require and display runtimeVersion in physical-device evidence; document stale-lock recovery and runtime identification.
- Disable implicit curl configuration in the iOS HTTP helper.

Validation: 126 mobile tests pass, including emitted Node competing-session behavior and the full iOS application retry regression. Forced root tests passed 26/26 tasks without cache; typecheck, lint, formatting, build, and docs checks passed. Physical iPhone/Android playback and execution of the mobile suite on Windows remain unverified locally. New hosted checks must be evaluated on this commit.

PR #302, opened 2026-09-01
fix(anidb): tolerate reindexed ids, extra languages, and 404 drift

KitsuneKode:
## Summary
Audited `anidb.app` against live site + current `ani-cli@5.0.2`. Live IDs reindexed (Solo Leveling `19413→4883`, `19837→4884`) and language inventory expanded (`kor` alongside `jpn`/`eng` on `16704`). Provider was surfacing retryable `network-error` for stale 404s and case-sensitive `jpn`/`eng` check.

## Changes
- `packages/providers/src/anidb/client.ts` `fetchAnidbEpisodes`/`fetchAnidbLanguages`: catch `HTTP 404` → `[]` so resolve surfaces `catalog-unavailable` not looping `network-error`. Adds live drift tolerance for reindexed slugs.
- `packages/providers/src/anidb/client.ts` `collectAnidbAvailableAudioModes`: `toLowerCase()==="jpn"/"eng"` (was case-sensitive) and explicit comment that `kor`/future codes are ignored — matches `languageEntryForMode` and keeps `sub/dub` only. Prevents `kor`-only episodes from incorrectly advertising `dub`.
- `packages/providers/src/anidb/direct.ts` `resolveAnidbShow`: when `providerNativeIds.anidb` present, probe episodes; if empty (404) fallback to `searchAnidb(title)` via `chooseAnidbSearchMatch`. Stale `sqlite` history entries (e.g. `19413`) auto-heal without manual re-search. Transient/Cloudflare errors keep direct (let caller surface `blocked`).

## Verification
- `bun x turbo typecheck --force` green (14 tasks)
- `bun x turbo test --filter=@kunai/providers --force` `544 pass`
- Live smoke (isolated profile `HOME=$(mktemp -d)`): `anidb-onigiri.smoke.ts` `ok true` `2 streams hls.anidb.app` `2 searchResults`, `chrome150` impersonate present, `mpv` available.
- `curl-impersonate` stays soft `degraded` in `anime` scope only (`dependency-rows.ts:100-119`, `install-commands.ts:143-147` already correct) — residential IP plain `curl` still works, CI/datacenter gets impersonate, relay remains metadata-only alternative. No installer hard-require change.

## Other providers
No immediate drift. `allmanga/miruro/vidrock/vidlink/videasy` etc share same `curl-impersonate` discovery; `hls.anidb.app` streams stay direct. `allmanga` mkissa chunk is highest future drift risk — recommend separate `verify:provider-parity` script, not in this PR.

Worktree: `.worktrees/fix-anidb-audit` `fix/anidb-audit-drift` `a5bc4648` (on top of `fix/ci-anidb-tls-impersonation` `91169f33`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved AniDB season matching for suffixed titles and reindexed show IDs.
  * Prevented valid empty catalogs from being mistaken for missing shows.
  * Improved handling of unavailable AniDB IDs and HTTP errors.
  * Recognized supported Japanese and English audio modes regardless of capitalization.
  * Ignored unsupported audio modes in live AniDB responses.

* **Documentation**
  * Updated the AniDB runtime contract with current matching, error-handling, and audio-mode behavior.

### DIFF
--- a/.docs/provider-dossiers/anidb-runtime-contract.md
+++ b/.docs/provider-dossiers/anidb-runtime-contract.md
@@ -1,6 +1,6 @@
 ---
 status: current
-lastReviewed: "2026-08-26"
+lastReviewed: "2026-09-01"
 ---
 
 # Provider: AniDB — runtime contract
@@ -52,7 +52,9 @@ instead of pretending those fields came from `anidb.app`.
 
 - `anidb.app` language evidence is per episode: `jpn` is the sub/original embed and `eng` is the
   dub embed when present. Search does not advertise both modes blindly; availability is confirmed
-  only by the episode languages response.
+  only by the episode languages response. Live `2026-09` also returns `kor` (e.g. Solo
+  Leveling S1 E1 `16704` → `eng,jpn,kor`); `kor` is ignored — `collectAnidbAvailableAudioModes`
+  remains `sub/dub` only (`jpn`/`eng` case-insensitive, exact).
 - A missing requested language is an exhausted AniDB attempt. It must not fall back to the other
   language and label the stream incorrectly.
 - Only exact `jpn` (sub/original) and `eng` (dub) catalog evidence is accepted.
@@ -92,7 +94,10 @@ AniDB models each season as its own title, so `routeAnidbSeason()` in
 - Season 1 retains the base title.
 - Season 2+ searches `<normalized base> Season N` and requires both a matching parsed season and
   exact/prefix normalized base-title evidence. An ambiguous best score fails closed with a
-  structured `not-found` rather than resolving the wrong title.
+  structured `not-found` rather than resolving the wrong title. A suffixed title such as
+  `Solo Leveling Season 2: Arise from the Shadow` (live `4884`, was `19837`) leaves
+  `solo leveling arise from the shadow` and matches via `prefix` (score 1) — covered by
+  `anidb-season-routing.test.ts` prefix test.
 - If no numbered sibling matches, a standalone `Final Season` may route to the ordinal immediately
   after the show's highest explicitly numbered sibling. Movie cards, final-season parts, and
   prefixed spin-off seasons provide no ordinal evidence; ambiguous or arc-named runs fail closed.
--- a/.docs/provider-dossiers/anidb-runtime-contract.md
+++ b/.docs/provider-dossiers/anidb-runtime-contract.md
@@ -57,6 +57,18 @@ instead of pretending those fields came from `anidb.app`.
   remains `sub/dub` only (`jpn`/`eng` case-insensitive, exact).
 - A missing requested language is an exhausted AniDB attempt. It must not fall back to the other
   language and label the stream incorrectly.
+- A reindexed show id (`19413` → `4883`) is repaired, but only on proof. `fetchAnidbEpisodeCatalog`
+  reports `missing: true` for an HTTP 404 and `missing: false` for a catalogue that is merely empty
+  or unparseable; `resolveAnidbShow` re-searches on the first and never on the second, because an
+  announced season with no episodes listed is not a stale id. The repair also requires real title
+  evidence (`chooseAnidbSearchMatch(..., { requireTitleEvidence: true })`) — the ordinary
+  "best guess is the top card" fallback is disabled there, since substituting a search result for a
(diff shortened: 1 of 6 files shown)


KitsuneKode (commented):
Reviewed against `.docs/agents/audit-findings-bar.md`. The language-case fix is a real bug correctly diagnosed. Three things to resolve first, one of them a regression trap hiding inside the merge conflict.

## 1. The conflict would delete #301's work — don't resolve it, drop the files

3 of the 7 files here already landed on `main` via #301, and **main's version is strictly better**:

| | this PR | `main` (#301) |
|---|---|---|
| `action.yml` | 60 lines | 128 lines |
| Platforms | Linux only | Linux + macOS + Windows (`pwsh` branch) |
| Digest pin | `sha256sum` only | `sha256sum` **or** `shasum` (macOS ships no `sha256sum`) |

Resolving the conflict by taking this branch's side silently drops Windows and macOS support from the live-provider lane. `.github/actions/setup-curl-impersonate/action.yml`, `.github/workflows/provider-matrix.yml` and `apps/cli/test/live/README.md` should be dropped from this branch entirely on rebase — main already has them, better. That leaves the 4 files this PR is actually about.

## 2. The two halves of the PR fight each other

This is the one I'd want changed before merge.

`client.ts` converts an explicit 404 into `[]`:

```ts
if (error instanceof Error && /HTTP 404/.test(error.message)) return [];
```

`direct.ts` then tries to recover the distinction that just got destroyed:

```ts
if (episodes.length === 0) { /* assume the id is stale, search by title */ }
```

But `[]` now means three different things: *this id is gone* (404), *this show has no episodes listed yet* (a just-announced season), and *the response parsed to nothing*. Only the first should trigger a title search. For the other two, `chooseAnidbSearchMatch(query, searched)` can return a **different show**, and we hand the user another title under the id they picked — the failure mode `routeAnidbSeason` deliberately fails closed to avoid.

The information needed for the right call exists at the 404 site and is thrown away one layer down, then guessed at one layer up. Suggest propagating it instead — a sentinel, a `status` on the error, or a `{ episodes, notFound }` shape — and gating the fallback on `notFound === true` rather than on `length === 0`.

## 3. No test covers any of the three behaviour changes

Checked both test dirs for all four changed functions before saying this:

```
resolveAnidbShow                   0 test files
fetchAnidbEpisodes                 0 test files
collectAnidbAvailableAudioModes    0 test files
fetchAnidbLanguages                0 test files
```

The test that *is* added covers `routeAnidbSeason` prefix matching — a function this PR does not change. It pins pre-existing behaviour, which is worth having, but it is not coverage for the fallback, the 404 handling, or the case fix. All three are stub-able at `anidbFetchText`, and `anidb.test.ts` already has the stub-URL harness.

Given CLAUDE.md's "declaration → reader" rule, #2 above is exactly the shape that needs a test to pin which meaning of `[]` triggers a re-search.

## Verified good — two things I specifically tried to disprove

- **The case fix is real.** `languageEntryForMode` (`client.ts:703`) already did `entry.code.toLowerCase() === code` while `collectAnidbAvailableAudioModes` did exact `=== "jpn"`. They genuinely disagreed, so `availableModes` could omit a mode the resolver would then happily find. Good catch, correctly justified.
- **The new `fetchAnidbEpisodes` in `resolveAnidbShow` is *not* the extra round-trip it looks like.** It looked like a new network call on every anime resolve, which would have been a hot-path regression. It isn't: `episodeCache` is a 30-minute TTL cache and `direct.ts:294` re-requests the same `showId` downstream, so this warms the cache rather than adding a fetch. Worth a comment saying so, or the next reader files it as a perf bug.

## Nits

- `catch { }` in `resolveAnidbShow` swallows aborts too; the comment only defends transient/Cloudflare. An aborted signal returning `direct` and continuing is not what the caller asked for.
- The 404 early-return sits before `episodeCache.set`, so a permanently-dead id re-404s on every call — and twice per resolve, at `direct.ts:108` and again at `:294`. Caching the negative would fix both.
- `/HTTP 404/.test(error.message)` reads a status out of prose. `anidbFetchText` throws it at `client.ts:162`; attaching `status` to the error is sturdier and would also give #2 its signal for free.

Happy to push the rebase-and-drop for the CI files if useful — that part is mechanical.

PR #306, opened 2026-09-01
Chromecast audo video

SirAlfie:
## What changed

  Adds Google Cast support across four areas:
  - Discovers compatible Google Cast devices on the local network and provides a playback-device
    picker.
  - Supports full video and audio casting with pause, resume, seeking, stopping, and playback
    progress controls.
  - Delivers header-protected media and converted WebVTT subtitles through secure, session-scoped LAN
    gateways.
  - Adds experimental --cast-audio playback, which plays muted video locally while casting extracted
    audio to a Chromecast with shared controls and clock correction.

  Regular --cast continues to use Google’s Default Media Receiver. Audio-only casting remains
  explicitly experimental and uses the separate --cast-audio option.

  ## Checklist

  - [x] Changeset added (bun run changeset). The casting changeset was folded into the staged 0.3.0
    package changelog, root changelog, and generated release artifact, then consumed according to the
    staged-release exception.

  - [x] bun run guard passes. @kitsunekode/kunai@0.3.0 is in sync.
  - [x] bun run fmt && bun run lint && bun run test && bun run typecheck passes locally. Tests
    requiring unavailable optional platform tools are skipped and remain enabled in supported CI
    environments.

  - [x] bun run build passes, including the CLI bundle and Linux host binary.
  - [x] Live provider and Discord smokes were skipped intentionally because this change does not
    modify provider resolution or Discord integration. Chromecast audio-only playback remains
    experimental.

## Summary by CodeRabbit

* **New Features**
  * Added experimental Google Cast playback with receiver discovery, device selection, remote controls, subtitles, and protected-stream support.
  * Added experimental audio-only casting, keeping video playback local while sending audio to a compatible receiver.
  * Added `--cast`, `--cast-audio`, and `--cast-list` options.
  * Added pause/resume and seeking with Space, arrow keys, and progress-bar clicks.

* **Documentation**
  * Updated CLI reference, keybindings, architecture documentation, feature maps, and release notes with casting details.


KitsuneKode (changes requested):
Thanks for taking on a difficult feature. The useful direction here is the receiver-neutral `PlaybackRouter`, explicit opt-in launch flags, opaque per-session gateway routes, and deterministic Cast protocol fixtures. That is a reasonable architecture to build on.

I am requesting changes because this is not merge-ready yet. This PR adds 4,468 lines across 77 files and touches playback lifecycle, network discovery, credential-bearing proxying, terminal input, release metadata, docs deployment, and a custom receiver in one change. The implementation currently has confirmed correctness/security failures, and the release/docs language promises behavior that has not been demonstrated on hardware.

## Additional blocking findings I verified

1. **[P1 security] Media credentials are forwarded to unrelated subtitle origins.** `google-cast-playback-backend.ts:146-158` passes `request.stream.headers` to the subtitle gateway for every provider and external subtitle track. `cast-subtitle-gateway.ts:82-89,133-136` then sends all of those headers to each track URL. Kunai can merge Wyzie/external subtitle URLs into `stream.subtitleList`, so a provider `Cookie` or `Authorization` header crosses origin. I reproduced this locally: an `https://unrelated.example/sub.srt` fetch received both `authorization: Bearer media-secret` and `cookie: session=secret`. Headers need an explicit per-resource/origin policy; media credentials must never be inherited by an unrelated subtitle URL. Add a regression test with a media origin and a different subtitle origin.

2. **[P1 correctness] Protected DASH is corrupted by the gateway.** `session-media-gateway.ts:125-141` treats DASH XML as regex-replaceable strings and replaces `SegmentTemplate` values with one opaque static route. For `media="chunk-$Number$.m4s"`, the receiver no longer sees `$Number$`; the registry instead stores a literal upstream URL containing `$Number$`. It also resolves the template against the MPD URL rather than the nested `BaseURL`. The current test at `session-media-gateway.test.ts:51-66` only asserts that the original strings disappeared, so it blesses the broken output. Either implement MPD-aware hierarchical URL/template rewriting (including `$Number$`, `$Time$`, nested `BaseURL`, `SegmentList`, and redirects) or remove protected DASH from the compatibility claim.

3. **[P1 lifecycle] A subtitle-gateway startup error leaks the already-open media gateway.** `google-cast-playback-backend.ts:126-160` starts and publishes the protected-media gateway, then awaits subtitle gateway startup before entering the `try/catch` at line 197. If LAN address selection or subtitle server binding fails, `play()` rejects while the tokenized media server remains listening. I reproduced the branch with a fake subtitle factory: media gateway close count remained `0`. Establish one ownership scope immediately after the first resource is acquired and unwind every partially constructed state.

4. **[P1 lifecycle] Split output announces ready/started before remote audio is usable.** `split-audio-playback-backend.ts:67-75` spreads the caller's `onPlaybackEvent` into local muted mpv and forwards `onPlayerReady` as soon as local video is ready. The remote receiver is only started at lines 119-180. `run-mpv-playback-session.ts:252-261` treats local `playback-started` as confirmed playback, acknowledging queue/history and starting presence before remote audio succeeds. A later Cast failure can therefore leave a failed split handoff recorded as playback. Suppress outward local lifecycle events during bootstrap and emit ready/started only when the composite session is actually usable; test receiver failure after local mpv starts.

5. **[P1 controls] Pause and seek are active during loading.** `playback-shell-input.ts:118-145` includes the player scope for non-playing operations, while `keybinding-runtime.ts:84-89` does not require `isPlaying` for pause/seek. That means Space/arrows can operate the local half of split output while it is deliberately paused waiting for remote audio, or a Cast control registered before `LOAD` completes. This can desynchronize startup. Gate transport controls to an established playable/composite state and add loading-state tests.

6. **[P1 docs/CI] The PR's docs gates fail.** `bun run verify:doc-frontmatter` fails because `.docs/casting.md` uses unsupported status `experimental` (allowed values are `current`, `draft`, `superseded`). `bun run --cwd apps/docs scripts/check-codegen-freshness.ts` also fails because `apps/docs/lib/generated-release-notes.json` was not regenerated after editing the 0.3.0 release notes. Both pass on the base revision / are introduced by this PR.

## Existing review threads that also need resolution

I independently checked the high-impact CodeRabbit findings and agree that they are substantive, not cosmetic:

- full A/V Cast is still blocked by the unconditional mpv dependency gate, even though that target does not use mpv;
- `ProgressBar` adds a second direct stdin reader beside Ink;
- transport commands start before `enqueueCommand`, so rapid pause/seek calls overlap (I reproduced two operations starting before the first completed);
- native Avahi discovery is unbounded and can extend `--cast-list` indefinitely;
- failed Cast `LAUNCH` leaks its TLS socket/heartbeat;
- repeat audio GETs create concurrent ffmpeg transcodes;
- router `stop()` can stop the currently selected backend instead of the backend that actually owns the active play;
- split output has no reliable cancellation path when local control acquisition fails;
- relative seeks use stale status instead of extrapolated position;
- manifest reads are unbounded when `Content-Length` is absent;
- nested/redirected media requests need origin-aware credential forwarding and revalidation.

The plaintext LAN bearer-token concern needs a documented threat-model decision rather than a reflexive “use HTTPS” change—Cast receivers often cannot trust an ephemeral local certificate—but the current docs should not call the route private without describing the trusted-LAN assumption, short lifetime, replay window, and mitigations.

## Codebase fit and quality

- Please split this into reviewable stages: (1) behavior-neutral router seam, (2) full A/V direct Cast + discovery, (3) protected media/subtitles, (4) shared controls, and (5) experimental split audio/custom receiver. Each stage should have independent acceptance evidence.
- Revert the unrelated `.oxfmtrc.json` policy weakening. Removing Tailwind sorting and globally excluding every README, CHANGELOG, test fixture, and docker-compose file is not part of Cast and reduces repository coverage.
- `KUNAI_EXPERIMENTAL_CAST` is stored in `AttentionFeatureFlags`, which is the wrong domain, and enabling it only registers a backend without selecting or exposing a target. That is effectively a user-visible no-op and conflicts with this repository's declaration-to-reader rule. Remove it or give it one documented, testable behavior.
- `runMpvPlaybackSession` now dispatches non-mpv backends. The underlying session policy can be shared, but the name and `PlayerOptions`-shaped boundary leave the abstraction half-completed.
- The changed-file anti-slop check reports 124 advisory findings, including unchecked assertions, open dictionaries, runtime `typeof` parsing, and `unknown` return values in the new Cast boundary. They are not all merge blockers individually, but they are a strong signal that the protocol/parser boundary needs typed validation rather than ad-hoc narrowing.

## Verification performed on this exact head (`c1742a1`)

- Focused Cast suite: **45 pass, 0 fail**. These are fixture/fake-session tests only.
- Forced root gates: typecheck **14/14**, lint **12/12**, format **26/26**, test **23/23 uncached**. CLI unit: **5,392 pass, 1 skip**; integration: **239 pass, 25 skip**.
- Build, host Linux binary, npm pack guard, release guard, release-notes check, PR build pipeline, and docs production build passed locally.
- `--cast-list` executes and exits cleanly on Linux with no receiver present.
- Docs frontmatter and generated-doc freshness fail as described above.
- No real Chromecast/Google TV, custom receiver app ID, deployed receiver, provider-protected stream, or Windows/macOS runtime was validated by those green tests.

## CI status and acceptance bar

The main CI and Release Guard runs did not execute any jobs. They are marked `action_required` because this is a public-fork PR from a first-time contributor and require maintainer approval. Vercel separately reports authorization required, so the custom receiver route is not deployed by this PR check. Until those workflows are approved, there is no Windows, macOS, docs-freshness, Linux binary, or hosted receiver evidence.

Before merge, please add regressions for the six findings above and the unresolved review threads, then provide manual hardware evidence for: discovery and direct IP; direct MP4/HLS/DASH; extensionless signed URLs; header-protected HLS and DASH; SRT/ASS/WebVTT subtitles including cross-origin tracks; pause/resume/repeated seeks/progress click/stop; receiver disconnect during connect, load, and playback; gateway/ffmpeg cleanup; split-audio startup, repeated seek, pause/resume, and 30-minute drift; and at least Linux plus the held Windows/macOS CI lanes. Full A/V and split-audio evidence should be reported separately.

The feature idea is worth pursuing, but the current patch overstates “works” and “safe.” The next revision should narrow the scope, make resource ownership and trust boundaries explicit, and make the tests assert outcomes rather than merely exercise branches.

KitsuneKode:
One process note for the next revision: please keep in touch with the existing CodeRabbit review as the branch changes. Work through each unresolved thread, reply with the exact fix and regression test (or concrete evidence when you disagree), and only resolve a thread after the updated code has been verified. After substantial pushes, re-run/request CodeRabbit so its findings are checked against the new head rather than silently becoming stale. Please also cross-reference those resolutions in your PR update; several of the current comments identify lifecycle, security, and cleanup defects that should be treated as part of the merge criteria, not optional bot feedback.

PR #311, opened 2026-09-02
test(diagnostics): stop the healthy-verdict test reading host memory

KitsuneKode:
## What flipped

`buildDiagnosticsInsight > healthy session reports healthy verdict and none action` failed during a pre-push full-suite run with no related change:

```
Expected: "healthy"
Received: "degraded"
```

It passes in isolation every time. The memory health row defaults to the **live process** — `process.memoryUsage()` plus `/proc` swap — and reports `warning` at RSS ≥ 3 GiB or any swap in use. Under 5,300+ tests in one Bun process on a machine with swap active, that condition is met, so a pure verdict test was asserting on the test runner's own memory.

## The change

- `BuildDiagnosticsInsightInput` gains `memorySnapshot?: RuntimeMemorySnapshot`, passed through to `buildRuntimeHealthSnapshot` exactly like the `memorySamples` it already accepted. Default behaviour is unchanged; production callers pass nothing and still read the live process.
- The healthy-session test injects a quiet snapshot.
- A new case drives the memory row to `degraded` through the same field, so the injection is a pinned seam rather than a way to mute the row.

## Why it matters for releases

The pre-push hook and CI run the whole suite. A test that reads host memory fails on whichever machine happens to be busiest — this one blocked pushing two release fixes today.

## Verification

13 pass in the file; forced `typecheck`, `lint`, `fmt` clean; the pre-push full suite passed on this push.

### DIFF
--- a/apps/cli/src/services/diagnostics/diagnostics-insight.ts
+++ b/apps/cli/src/services/diagnostics/diagnostics-insight.ts
@@ -10,7 +10,7 @@ import type { ProviderHealth, ProviderId } from "@kunai/types";
 
 import type { DiagnosticEvent } from "./diagnostic-event";
 import { buildRuntimeHealthSnapshot } from "./runtime-health";
-import type { RuntimeMemorySample } from "./runtime-memory";
+import type { RuntimeMemorySample, RuntimeMemorySnapshot } from "./runtime-memory";
 
 // ---------------------------------------------------------------------------
 // Shared taxonomy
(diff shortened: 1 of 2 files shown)


KitsuneKode:
**Merge this after 0.3.0 is published, not before.**

This PR changes `apps/cli/src/services/diagnostics/diagnostics-insight.ts`, which is compiled into the release binaries and therefore into all eight platform-package tarballs. Three of those are already live at 0.3.0 (`linux-x64`, `linux-x64-musl`, `linux-arm64`) with the bytes from `ba66378`. The publisher skips a live version only on identical integrity and halts on a different one, so merging any `apps/cli/src` or `packages/` change before the 0.3.0 rerun would force a version bump.

#310 and #312 touch only `scripts/`, workflows, docs, and tests, so a candidate rebuilt on top of them stays byte-identical for the three live packages.

PR #326, opened 2026-09-02
fix(anime): stop a stale relay, an AniList outage, and a broken fallback cycle blanking the anime lane

KitsuneKode:
## Why

Every anime provider went dark. It looked like a Kunai regression; it was not. Three external conditions plus one client bug that turned each of them from recoverable into total.

The load-bearing one: a relay deployed before a provider exists answers its RPC route with its own `404 unknown-provider`. The client read that status as the *upstream's* verdict, so AniDB reported "no such anime", cached the miss, and took the whole anime lane down — while the same ids resolved fine over curl one layer below.

## What changed

**Relay-hop statuses are no longer read as upstream verdicts.** Relayed responses now carry `X-Kunai-Relay-Hop`, written by the client port after the fact and always overwriting, so an upstream can neither forge nor suppress it. AniDB only treats a 404 as authoritative when it did not arrive over a hop.

Note the rejected approach: reading the relay's JSON error body to decide fallback breaks a deliberate security contract — a forged error envelope would downgrade a user off the relay and expose their IP. `packages/relay/test/relay-fetch-port.test.ts` already locked that, and the fix respects it.

**`verify:deploy` can now see the drift it exists to catch.** It probed a hardcoded list of the five providers that existed when it was written and only failed on `host-not-allowed`, so it passed against a relay that had no `anidb` at all. It now derives one probe per registered provider and fails on `unknown-provider`, plus on an auth wall that was silently reporting an unverified registry as healthy.

**Anime search survives an AniList outage.** AniList answers 403 (`"temporarily disabled due to severe stability issues"`). Miruro carries no native search, so that registry service was its only catalog route and its search box went dead — while AniDB and AllManga could still answer from their own catalogs. Anime search now borrows native search from a sibling anime provider before giving up.

**Cross-catalog identity reaches provider fallbacks.** AniDB answers search with `gintama-1816` and only learns the AniList/MAL ids later, in `listEpisodes` — and `EpisodeIdentity` has nowhere to carry them, so they died on the episode rows. Falling back to Miruro handed over a title with no AniList id and was rejected `unsupported-title`. With AllAnime gated upstream that left the *default* anime provider with no reachable fallback at all: the cycle looked healthy and could never advance. AllManga never hit this because AllAnime's search response includes `aniListId`.

Ids discovered while listing episodes are now lifted onto the title as the resolve input is built — before lane adaptation, which also gates on holding an anime catalog id. Gap-filling only, so a stale episode row cannot rewrite a title that already knows its identity.

**Quitting mid-resolve is a cancellation, not a network fault.** Ctrl-C kills the curl child with exit 130, whose message matched neither `"abort"` nor `"cancel"`. A plain quit was classified as an unknown, retryable network fault: it spent a fallback provider attempt during shutdown and logged an ERROR for a user who simply quit.

## Verification

Gates run with the turbo cache bypassed, per `CLAUDE.md`: **23/23 test tasks (0 fail)**, typecheck 14/14, lint 0 errors, `fmt:check`, `verify:doc-paths`. Each new test was confirmed red without its fix.

Live, against real providers:

| Check | Before | After |
|---|---|---|
| AniDB via a stale relay | `No AniDB streams` | 2 reachable streams |
| Miruro via relay | — | resolves |
| Fresh install, no relay — movie / series / anime | — | all three resolve, reachable |

The relay behind the original outage was redeployed (5 → 6 providers) — a relay is user-owned, so the client fix is what stops this reaching anyone else.

## Not in scope

- **AllAnime is dead upstream**, not fixable here: `NEED_CAPTCHA` with *or without* valid crypto direct, `PersistedQueryNotFound` via relay, and bootstrap rejects our pinned build 140 as retired. ani-cli deleted its AllAnime path on 2026-08-01 and moved to `anidb.app`. Demoting it out of the default anime priority is a product call, left alone here.
- **`canHandle` is still loose** — Miruro can be offered titles it structurally cannot use. This PR means it usually now gets a usable id; tightening the guard deserves its own change.

### DIFF
--- a/apps/cli/src/services/providers/stream-request-adapter.ts
+++ b/apps/cli/src/services/providers/stream-request-adapter.ts
@@ -41,10 +41,14 @@ export function streamRequestToResolveInput(
   options?: StreamRequestAdapterOptions,
 ): ProviderResolveInput {
   assertTitleMatchesShellMode(request.title, mode);
+  // Enriched once, before lane adaptation: `adaptResolveLane` gates the
+  // anime lane on holding an anime catalog id, so a title that only learns its
+  // AniList id from the episode rows has to carry it by this point too.
+  const title = withEpisodeCatalogIdentity(request.title, request.episode);
   const naturalKind: MediaKind =
-    mode === "youtube" ? "video" : mode === "anime" ? "anime" : request.title.type;
+    mode === "youtube" ? "video" : mode === "anime" ? "anime" : title.type;
   const adapted = adaptResolveLane({
-    title: request.title,
+    title,
     naturalKind,
     episode: episodeToCoreIdentity(request.episode),
     catalogIdentity,
@@ -54,7 +58,7 @@ export function streamRequestToResolveInput(
   const animeAudioIntent =
     adapted.mediaKind === "anime" ? resolveAnimeAudioIntent(request.audioPreference) : null;
   return {
-    title: titleToCoreIdentity(request.title, mode, catalogIdentity, providerId, adapted.mediaKind),
+    title: titleToCoreIdentity(title, mode, catalogIdentity, providerId, adapted.mediaKind),
     episode: adapted.episode,
     mediaKind: adapted.mediaKind,
     preferredSourceId: normalizeOptionalId(request.selectedSourceId),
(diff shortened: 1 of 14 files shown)


KitsuneKode:
## Review & Approval Recommendation

Conducted a full architecture and security review of this PR:

1. **Relay Hop Status vs. Upstream Miss**:
   - In `packages/relay/src/create-relay-fetch-port.ts`, stamping `X-Kunai-Relay-Hop: 1` on relayed responses cleanly stops stale relay 404s (`unknown-provider`) from being mistaken for authoritative upstream catalog misses.
   - Client-side stamping ensures upstreams cannot spoof or suppress the marker.
2. **AniList Outage Resilience**:
   - Search routing fallback to sibling anime providers (AniDB, AllManga) with native search when AniList returns 403 or 5xx allows the anime catalog to survive third-party outages. Cancellation signals are properly honored.
3. **Cross-Catalog Identity Lifting**:
   - In `stream-request-adapter.ts`, discovered AniList/MAL IDs are lifted to `title.externalIds` as non-destructive gap-filling (`??`), ensuring fallback providers like Miruro receive necessary catalog IDs without risk of rewriting established titles.
4. **Cancellation vs. Network Error**:
   - Exit codes 130 and 143 from curl subprocesses are classified as user cancellations rather than triggering retries and spurious error logs.

### Merge Order & Rebase Note
- **Merge this PR first**, followed by stacked PR #329.
- **Attention during PR #345 rebase**: PR #345's refactoring in `packages/providers/src/anidb/client.ts` touches `anidbFetchText`. Ensure that PR #345 preserves the `RELAY_HOP_HEADER` check so stale relays do not throw `AnidbHttpStatusError(404)` and resurrect the anime lane outage.

Issue #319, opened 2026-09-02
Test suite is flaky because timing assertions race the parallel runner


KitsuneKode:
## What happens

Full-suite runs fail on a *different* test each time, and none of them reproduce in
isolation. Three consecutive local runs of `bun run test:unit`:

```
run A:  (fail) diagnostics-panel-lines > renders sections in cockpit order
run B:  (fail) resolveProviderStreamWithRetries > classifies provider timeouts…
run C:  0 fail
```

Each of those passes 5/5 when run alone.

## Why

Not mocking pollution — only 2 test files call `mock.module`, and neither is near the
affected areas.

The cause is tests asserting on **real wall-clock timing with margins smaller than the
scheduling noise of a saturated parallel suite**. The clearest case, from run B
(`apps/cli/test/unit/app/provider-resolve-retry.test.ts:73`):

```ts
timeoutMs: 1,                                   // product must time out in 1ms
resolve: async () => {
  await new Promise((r) => setTimeout(r, 20));  // work takes 20ms
```

A 19ms margin. Starve the event loop past that under Turbo + Bun parallelism and the
resolve wins, so the expected timeout never fires. Whichever timing-sensitive test
loses the scheduler lottery on a given run is the one that fails — exactly the
"different test every time" signature.

## Scale

- **32** test files use a real `Bun.sleep`
- **18** build deadlines from `Date.now() + N`
- `apps/cli/test/unit/architecture/test-timeout-budget.test.ts` already records one
  storage test measured at **6428ms on one run and 16549ms on the next** — a 2.6x
  spread on the same machine

## Instances already patched, one at a time

Each was diagnosed and fixed individually, which is the cost of not fixing the class:

| PR | Guard | Failed at |
|---|---|---|
| #312 | installer harness 5s hang guard | loaded runner, concurrent downloads |
| #311 | diagnostics read the runner's own RSS/swap | host memory, not the product |
| (earlier) | `RETAIN_TEST_TIMEOUT_MS` 40ms | 50.57ms on Windows |
| #318 | `RECLAIM_TEST_TIMEOUT_MS` 100ms | **285.92ms** |
| #318 | `install.ps1` zip test, shared 20s budget | **20000.70ms** — 0.7ms over |

## What to do

`AGENTS.md` already states the rule — *"A test that needs a timeout or a real sleep to
pass is wrong. Anchor injected clocks past the data they read"* — the suite just does
not follow it yet.

[...]

PR #329, opened 2026-09-02
fix(relay,providers): forward VidLink's playback header, terminate curl argv, correct the recovery key

KitsuneKode:
Stacked on #326 — that PR touches the same `anidbFetchText` argv this one changes, so branching from `main` would guarantee a conflict. **Merge #326 first**; the diff here is only the three commits' worth of changes below.

Three findings from an audit sweep, kept because they survived verification against `.docs/agents/audit-findings-bar.md`. Several other findings in the same sweep were refuted and are deliberately not here.

## VidLink is unplayable through a relay

VidLink asks for the browser playback path with `x-playback-environment: webkit` (`vidlink/direct.ts:200`). The relay's metadata allowlist never carried it, so a relayed VidLink lane got `deliveryType: "file"` — bcdn MP4s flagged `requiresProxy` that answer 429 to every non-browser client. It resolves, then will not play, with the relay as the only difference from a working direct request.

`vidlink-dash.test.ts` only asserted that the provider *sends* the header on the direct path; nothing pinned the allowlist, so this was invisible to the suite. Now covered both ways — forwarded, and credentials still dropped.

**Relay operators must redeploy.** The allowlist is baked in at deploy time, so an already-deployed relay keeps stripping the header.

## `--` before the URL operand in both curl call sites

AniDB and Miruro both put the fetch URL last in argv, and those URLs come from upstream JSON (`embed_url`, the Miruro pipe payload). Without a terminator, curl reads a value beginning with `-` as options rather than as the address, so upstream JSON gets a say in a local subprocess's argv.

Scoped deliberately. This is hardening, not an incident: a single trailing argv slot cannot land a working option — `curl … "-o/tmp/pwned"` exits 2 with `no URL specified` and writes nothing — and the whole scenario presupposes a hostile anidb.app, which can already hand mpv arbitrary URLs.

**A scheme allowlist was considered and rejected here.** Rejecting protocol-relative or relative embed URLs would regress the anime lane, which is a likelier outcome than the injection it prevents. `--` closes the option half completely at zero behavioural cost; the scheme check deserves its own change with anime-lane care.

Guarded at the source, in the style of `boundary-imports.test.ts`, because the argv is assembled inside the fetch helpers with no seam to observe it from a behavioural test. Both guards were confirmed red with the terminator removed.

## README documents a recovery key that does nothing

In-player recovery is `Ctrl+R` in both `app-shell/keybindings.ts:514` and `assets/mpv/kunai-bridge.lua:1085`. `README.md` said a bare `r` in two places. `.docs/keybindings.md` was already correct.

Generating that table from `KEYBINDINGS` is already filed in `.plans/023-cli-surface-honesty.md`; this is the textual half only.

## Verification

Full suite with the turbo cache bypassed: **23/23 tasks, 0 fail**. Typecheck 14/14, lint 0 errors, `fmt:check` 26/26, `verify:doc-paths`.

Worth flagging separately: the suite failed twice during this work and passed on re-run with no changes, in `@kunai/providers` and in the pre-push hook. That is the flakiness tracked in #319/#323, not this branch — but it does mean a red run here should be re-run before it is believed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -433,7 +433,7 @@ page or in-app `?` over duplicating chords here.
 ### Playback
 
 - Streams are resolved from direct-provider sources and handed to `mpv`.
-- **Recover** (`r`) refreshes the current stream and resumes from last position.
+- **Recover** (`Ctrl+R`) refreshes the current stream and resumes from last position.
 - **Recompute sources** (`/recompute`) bypasses cached provider memory when provider state looks stale.
 - **Fallback** (`⇧F`) tries the next compatible provider when the current one fails.
 - **Source / quality picker** switches among already-resolved stream options.
@@ -633,7 +633,7 @@ not ship as runtime behavior.
 <details>
 <summary><b>Search works but playback fails or stalls.</b></summary>
 
-Providers break when upstream sites change. In playback, press `r` to recover the
+Providers break when upstream sites change. In playback, press `Ctrl+R` to recover the
 stream, `⇧F` to fall back to the next compatible provider, `o` to pick another
 source, or `k` to pick quality. If sources look stale, `/recompute` bypasses cached provider
 memory. Persistent issues → `/diagnostics`, then `/export-diagnostics` for a
--- a/packages/providers/src/anidb/client.ts
+++ b/packages/providers/src/anidb/client.ts
@@ -248,6 +248,10 @@ export async function anidbFetchText(
     maxTime,
     ...anidbCipherArgs(curl.impersonates),
     ...(options.reportStatus === true ? ANIDB_STATUS_WRITE_OUT : []),
+    // Everything after `--` is an operand, never an option. Embed and playlist
+    // URLs arrive from upstream JSON, so without this a value beginning with
+    // `-` would be read as curl flags rather than as the address to fetch.
+    "--",
     url,
   ];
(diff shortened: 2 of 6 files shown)


KitsuneKode:
## Review & Approval Recommendation

Conducted a security and runtime audit of the changes in this PR:

1. **VidLink Header Forwarding**:
   - Adding `x-playback-environment` to `METADATA_HEADER_ALLOWLIST` fixes the BunnyCDN 429 issue over relay connections.
   - Authorization headers, session cookies, and hop-by-hop headers remain strictly stripped by `filterForwardHeaders`.
   - Header values are guarded against CRLF injection via `containsInvalidHeaderText`.
2. **Curl Argv Option Termination**:
   - Prepending `"--"` before target URLs in AniDB and Miruro fetchers guarantees that upstream JSON values starting with `-` cannot land as subprocess options in local curl executions.
   - Verified by `curl-argv-terminator.test.ts`.
3. **Keybinding Documentation**:
   - Corrected in-player recovery key from `r` to `Ctrl+R` in `README.md`, matching `kunai-bridge.lua` and `.docs/keybindings.md`.

### Merge Recommendation
This PR is cleanly stacked on #326 and ready to merge immediately after #326 is merged.

PR #333, opened 2026-09-02
fix(install): provision yt-dlp and curl-impersonate on Windows

KitsuneKode:
## What changed

Windows `irm … | iex` now drops a verified portable `yt-dlp` and (x64) curl-impersonate into `%LOCALAPPDATA%\kunai\deps\` so YouTube and anime search work after a one-click install, instead of leaving doctor warnings and a `winget install yt-dlp` command that cannot actually run.

## Why

After a native Windows install, Kunai showed:

- **yt-dlp missing** — doctor printed `winget install yt-dlp`, which matches both `yt-dlp.yt-dlp` and a Microsoft Store listing and exits with "Multiple packages found".
- **curl-impersonate missing** — Cloudflare fingerprints the TLS handshake, so anime search can come back empty. There is no winget/scoop/choco package; doctor only linked the GitHub releases page.

`irm … | iex` also cannot confirm winget licence agreements, so the installer previously *printed* those commands and never installed anything. mpv is unchanged: it is still a package-manager prompt.

## What the installer does now

On the **binary** Windows path, unless `-SkipDeps` / `KUNAI_SKIP_DEPS=1`:

1. Download official `yt-dlp.exe` (or `yt-dlp_arm64.exe`) from GitHub, verify `SHA2-256SUMS`, install under `%LOCALAPPDATA%\kunai\deps\yt-dlp`, put that folder on the User PATH.
2. Download the same curl-impersonate x64 archive CI pins (`v2.2.1`), verify the digest, extract, PATH the directory that contains `curl_chrome*.bat`. ARM64 is warned: upstream has no Windows ARM64 build.
3. Failures here **do not fail** the Kunai install; the exact fallback command is printed (`winget install --id yt-dlp.yt-dlp -e`, or the curl-impersonate releases URL).

`mpv` remains opt-in via winget/scoop/choco. Doctor/setup copy-paste is now `winget install --id yt-dlp.yt-dlp -e`.

## Testing

Fixture-backed install runs `Install-PortableYtDlp` and `Install-PortableCurlImpersonate` against a localhost GitHub stand-in:

- verified `yt-dlp.exe` plus a `curl_chrome*.bat` tarball land under `KUNAI_DATA_DIR/deps` (not under a nested `extract` folder)
- checksum mismatch leaves no dest and prints `winget install --id yt-dlp.yt-dlp -e`
- already-present dest files skip the download
- a leftover dest directory is replaced in place
- a locked dest directory fails closed instead of nesting extract (POSIX; Windows ACLs are not `chmod`)

Wrapper-dir discovery compares `realpath` of both sides (Windows `RUNNER~1` vs `runneradmin`, macOS `/var` vs `/private/var`).

## CodeRabbit

Acted on all three notes:

- changeset uses `-SkipDeps`, not Bash `--skip-deps`
- install doc no longer claims two portable tools on every Windows host
- `Remove-Item` failure no longer lets `Move-Item` nest `extract` under dest

Skipped the docstring-coverage warning: that metric is JSDoc on touched functions, not `install.ps1` comment blocks.

## Checklist

- [x] Changeset added (`bun run changeset`) — required for user-facing CLI changes; N/A for docs-only or non-release infra
- [x] `bun run guard` passes when `apps/cli/package.json`, changelogs, or `.changeset/**` changed
- [x] `bun run fmt && bun run lint && bun run test && bun run typecheck` passes locally
- [ ] `bun run build` passes for feature, playback, provider, release, or packaging-sensitive changes — installer/docs/script change; no binary rebuild
- [x] Live provider or Discord smokes were skipped intentionally, or run manually with results noted

## Seams

- **Declaration → reader:** `KUNAI_YTDLP_RELEASE_BASE`, `KUNAI_CURL_IMPERSONATE_VERSION` / `_SHA256` / `_RELEASE_BASE` are read by the portable installers; `-SkipDeps` still skips them.
- **Reverse states:** already-present PATH copies are left alone; `--skip-deps` / `-SkipDeps` is the way out; a locked dest fails closed; `kunai uninstall --purge` removes `%LOCALAPPDATA%\kunai` including `deps\`.
- **Entry points:** installer (binary method) plus doctor/setup/README fallbacks. npm/bun/source installs are unchanged (no silent winget).
- **Both lanes:** yt-dlp is the YouTube/download lane; curl-impersonate is the anime Cloudflare lane.
- **Every provider:** runtime tools, not adapter-specific. AllManga/AniDB/Miruro keep discovering `curl_chrome*.bat` from PATH.
- **Every platform:** Windows-only portable install. Wrapper-dir identity and dest-replace tests avoid pinning one OS's path or ACL spelling. Linux/macOS installs still print the package-manager command and wait for `[y/N]`.
- **Docs:** `docs/users/install-and-update.mdx`, `.docs/quickstart.md`, `.docs/debugging-map.md`, `.docs/testing-strategy.md`, root and CLI READMEs.

<div><a href="https://cursor.com/agents/bc-01a06468-cf59-7ae6-ab1c-3e5f53837158?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-01a06468-cf59-7ae6-ab1c-3e5f53837158&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

## Summary by CodeRabbit

* **New Features**
  * Windows installation now provisions verified portable yt-dlp and, on x64 systems, curl-impersonate.
  * Helpers are stored in Kunai’s data directory, added to the user PATH, and repaired when missing or modified.
  * `--skip-deps` can skip portable helpers and package-manager prompts.
  * Windows yt-dlp installation guidance now uses the exact WinGet package ID.

* **Bug Fixes**
  * HLS playback on Windows no longer fails when the installed curl lacks HTTP/2 support.
  * The curl upgrade prompt remains available when curl-impersonate is installed.

* **Documentation**
  * Updated Windows installation, troubleshooting, and dependency guidance.

### DIFF
--- a/apps/cli/test/integration/install-scripts-pwsh.test.ts
+++ b/apps/cli/test/integration/install-scripts-pwsh.test.ts
@@ -32,6 +32,7 @@ import {
 
 const REPO_ROOT = join(import.meta.dirname, "../../../..");
 const INSTALL_PS1 = join(REPO_ROOT, "install.ps1");
+const PWSH_PATH = Bun.which("pwsh") ?? "pwsh";
 
 function readInstallerManifest(configDir: string): InstallManifest {
   return JSON.parse(readFileSync(join(configDir, "install.json"), "utf8"));
@@ -2177,8 +2178,11 @@ async function runPortableHelperProbe(
     "Write-DownloadProgress",
     "Clear-DownloadProgress",
     "Invoke-BoundedDownload",
+    "Write-Utf8File",
     "Get-YtdlpReleaseAsset",
     "Get-HelperChecksumEntry",
+    "Test-ManagedFileDigest",
+    "Test-ManagedSourceDigest",
     "Get-CurlImpersonateWindowsSpec",
     "Test-CurlImpersonatePresent",
     "Find-CurlImpersonateWrapperDir",
@@ -2218,7 +2222,9 @@ async function runPortableHelperProbe(
       "$script:LastDownloadHttpStatus = $null",
       'function Write-Info($m) { Write-Host "-> $m" }',
       'function Write-Warn($m) { Write-Host "! $m" }',
-      "function Test-Cmd($name) { [bool](Get-Command $name -ErrorAction SilentlyContinue) }",
+      // The fixture owns yt-dlp state. Never let a host installation turn a
+      // download/repair assertion into an accidental already-present branch.
+      "function Test-Cmd($name) { if ($name -eq 'yt-dlp') { return $false }; [bool](Get-Command $name -ErrorAction SilentlyContinue) }",
       "function Test-RetryableHttpStatus([int]$Status) { return ($Status -eq 408 -or $Status -eq 429 -or $Status -ge 500) }",
       "function Get-WindowsArch { return 'x64' }",
       // Stands in for the -SkipPathUpdate branch of the real Add-UserPath: it
@@ -2231,7 +2237,7 @@ async function runPortableHelperProbe(
     ].join("\n"),
   );
 
-  const proc = Bun.spawn(["pwsh", "-NoProfile", "-File", probePath], {
+  const proc = Bun.spawn([PWSH_PATH, "-NoProfile", "-File", probePath], {
     env: { ...BOUNDED_DOWNLOAD_ENV, ...env },
     stdout: "pipe",
     stderr: "pipe",
@@ -2385,6 +2391,7 @@ describePwsh("install.ps1 portable Windows helpers", () => {
       const dir = mkdtempSync(join(tmpdir(), "kunai-wrapper-family-"));
       try {
         writeFileSync(join(dir, file), "@echo off\r\n");
+        writeFileSync(join(dir, "curl-impersonate.exe"), "MZ\n");
         const output = evalHelperFns(`
           $env:Path = ${JSON.stringify(dir)}
           if (Test-CurlImpersonatePresent) { Write-Output 'present' } else { Write-Output 'missing' }
@@ -2503,14 +2510,31 @@ describePwsh("install.ps1 portable Windows helpers", () => {
     }
   });
 
-  test("already-present dest files skip the GitHub download", async () => {
+  test("verified already-present dest files skip the GitHub download", async () => {
     const sandbox = createInstallerSandbox("install-ps1-portable-helpers-present");
     mkdirSync(join(sandbox.dataDir, "deps", "yt-dlp"), { recursive: true });
     mkdirSync(join(sandbox.dataDir, "deps", "curl-impersonate"), { recursive: true });
-    writeFileSync(join(sandbox.dataDir, "deps", "yt-dlp", "yt-dlp.exe"), "MZ-already-present\n");
+    const ytdlpBody = "MZ-already-present\n";
+    const ytdlpDigest = createHash("sha256").update(ytdlpBody).digest("hex");
+    writeFileSync(join(sandbox.dataDir, "deps", "yt-dlp", "yt-dlp.exe"), ytdlpBody);
+    writeFileSync(join(sandbox.dataDir, "deps", "yt-dlp", "yt-dlp.exe.sha256"), `${ytdlpDigest}\n`);
     writeFileSync(
       join(sandbox.dataDir, "deps", "curl-impersonate", "curl_chrome146.bat"),
       "@echo off\r\n",
+    );
+    const backendBody = "MZ-curl-impersonate\n";
+    const backendDigest = createHash("sha256").update(backendBody).digest("hex");
+    writeFileSync(
+      join(sandbox.dataDir, "deps", "curl-impersonate", "curl-impersonate.exe"),
+      backendBody,
+    );
+    writeFileSync(
+      join(sandbox.dataDir, "deps", "curl-impersonate", ".kunai-source.sha256"),
+      `${"f7faa8c42b63b4a96245429e46956e11ae7d7076d60f65768c0018d3bb18d7e5"}\n`,
+    );
+    writeFileSync(
+      join(sandbox.dataDir, "deps", "curl-impersonate", ".kunai-backend.sha256"),
+      `${backendDigest}\n`,
     );
     try {
       await withReleaseFixture({}, async (_baseUrl, evidence) => {
@@ -2532,6 +2556,54 @@ describePwsh("install.ps1 portable Windows helpers", () => {
         expect(result.stdout).toContain("curl-impersonate already present");
         expect(evidence.requests).toEqual([]);
       });
+    } finally {
+      sandbox.cleanup();
+    }
+  });
+
+  test("repairs unverified managed helpers instead of trusting their filenames", async () => {
+    const sandbox = createInstallerSandbox("install-ps1-portable-helpers-repair");
+    const ytdlpDir = join(sandbox.dataDir, "deps", "yt-dlp");
+    const impersonateDir = join(sandbox.dataDir, "deps", "curl-impersonate");
+    mkdirSync(ytdlpDir, { recursive: true });
+    mkdirSync(impersonateDir, { recursive: true });
+    writeFileSync(join(ytdlpDir, "yt-dlp.exe"), "");
+    writeFileSync(join(impersonateDir, "curl_chrome146.bat"), "@echo off\r\n");
+
+    const ytdlpBody = "MZ-repaired-yt-dlp\n";
+    const ytdlpDigest = createHash("sha256").update(ytdlpBody).digest("hex");
+    const archive = createCurlImpersonateArchive(sandbox.root);
+    try {
+      await withReleaseFixture(
+        {
+          "/SHA2-256SUMS": { body: `${ytdlpDigest}  yt-dlp.exe\n` },
+          "/yt-dlp.exe": { body: ytdlpBody },
+          [`/${archive.asset}`]: { body: archive.bytes },
+        },
+        async (baseUrl, evidence) => {
+          const result = await runPortableHelperProbe(
+            {
+              ...isolatedHelperEnv(sandbox),
+              KUNAI_YTDLP_RELEASE_BASE: baseUrl,
+              KUNAI_CURL_IMPERSONATE_RELEASE_BASE: baseUrl,
+              KUNAI_CURL_IMPERSONATE_SHA256: archive.sha256,
+            },
+            [
+              "$ytdlp = [bool](Install-PortableYtDlp)",
+              "$imp = [bool](Install-PortableCurlImpersonate)",
+              'Write-Output "RESULT ytdlp=$ytdlp impersonate=$imp"',
+            ].join("\n"),
+          );
+
+          expect(result.status, `${result.stdout}${result.stderr}`).toBe(0);
+          expect(result.stdout).toContain("RESULT ytdlp=True impersonate=True");
+          expect(readFileSync(join(ytdlpDir, "yt-dlp.exe"), "utf8")).toBe(ytdlpBody);
+          expect(existsSync(join(impersonateDir, "curl-impersonate.exe"))).toBe(true);
+          expect(evidence.requests).toContain("/SHA2-256SUMS");
+          expect(evidence.requests).toContain("/yt-dlp.exe");
+          expect(evidence.requests).toContain(`/${archive.asset}`);
+        },
+      );
     } finally {
       sandbox.cleanup();
     }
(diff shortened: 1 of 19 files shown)


install.ps1
```diff
@@ -1616,25 +1907,30 @@ function Install-OptionalDeps {
...
   # against that build, so offer the full winget/scoop curl when the one on PATH
-  # cannot do HTTP/2.
+  # cannot do HTTP/2. An impersonate build already covers that (and Cloudflare),
+  # so skip the upgrade when we just installed one or found one.
   $curlNeedsUpgrade = $false
-  if (Test-Cmd 'curl') {
-    try {
-      $curlFeatures = (& curl.exe --version 2>$null) -join "`n"
-      $curlNeedsUpgrade = -not ($curlFeatures -match 'HTTP2')
+  if (-not $portableImpersonate -and -not (Test-CurlImpersonatePresent)) {
```
KitsuneKode: **This suppresses the HTTP/2 curl remedy for two call sites that never go through `resolveCurlCandidate`.**

The comment above says an impersonate build "already covers that (and Cloudflare)". That holds for AniDB and Miruro's fetch path, which do resolve through `resolveCurlCandidate`. It does not hold here:

- `apps/cli/src/infra/player/hls-relay.ts:213` spawns the literal string `"curl"`, and `buildHlsRelayCurlArgs` (`hls-relay.ts:187`) passes `--http2` unconditionally. `startHlsRelay` gates only on `Bun.which("curl")` (`hls-relay.ts:384`). Against the Schannel `curl.exe`, `--http2` exits with *"the installed libcurl version doesn't support this"*, so the relay fails outright for every stream `streamNeedsHlsRelay` matches. The call site (`PlayerServiceImpl.ts:862`) has no platform gate.
- `packages/providers/src/miruro/direct.ts:102` `detectCurlHttp2Support()` probes plain `curl --version`, so Windows drops `--http2` even while invoking the impersonate binary. That mismatch predates this PR, but the prompt being suppressed here was the thing that fixed it.

I pulled the pinned archive to check whether PATHing the deps directory incidentally supplies an HTTP/2 curl. It does not — `curl-impersonate-v2.2.1.x86_64-win32.tar.gz` contains `curl-impersonate.exe`, 34 `curl_*.bat` wrappers, and two LICENSE files. No `curl.exe`.

So on the new default happy path (impersonate installs successfully) the remedy disappears for a failure mode that still exists. Suggest leaving the HTTP/2 check unconditional and only softening the *wording* when an impersonate build is present — or routing `hls-relay` through `resolveCurlCandidate` first, though that is a larger change than this PR wants.

---
_Generated by [Claude Code](https://claude.ai/code)_

install.ps1
```diff
@@ -1616,25 +1907,30 @@ function Install-OptionalDeps {
...
-    $curlNeedsUpgrade = $true
   }
   if ($curlNeedsUpgrade) {
     Write-Warn 'The curl on PATH has no HTTP/2 support (Windows ships a Schannel build).'
     $missing += 'curl'
   }
 
   if ($missing.Count -eq 0) {
-    Write-Info 'mpv, yt-dlp and curl are already installed.'
+    if (Test-Cmd 'mpv') {
```
KitsuneKode: **Dead guard.** A missing mpv is pushed into `$missing` a few lines above, so reaching `$missing.Count -eq 0` already implies mpv is on PATH. `Test-Cmd 'mpv'` is always true here.

---
_Generated by [Claude Code](https://claude.ai/code)_

install.ps1
```diff
@@ -1594,20 +1619,286 @@ function Request-OptionalInstall {
...
 }
 
+# Session PATH plus the persistent User PATH. SkipPathUpdate still skips the
+# registry write (managed/test environments own PATH), but the current process
+# has to see a helper we just dropped or the rest of this install cannot find it.
+function Register-HelperPath([string]$Dir) {
+  if ([string]::IsNullOrWhiteSpace($Dir)) { return }
+  $normalized = $Dir.Trim().Trim('"').TrimEnd('\')
+  $entries = @($env:Path -split ';' | ForEach-Object { $_.Trim().Trim('"').TrimEnd('\') })
+  if ($entries -notcontains $normalized) {
```
KitsuneKode, line 1634: Minor: this prepends to `$env:Path`, then `Add-UserPath` prepends the same directory again whenever it is not already in the **User** PATH — so the session PATH ends up with a duplicate entry on a first install. Cosmetic only, but it is avoidable by letting `Add-UserPath` own the session write and only doing it here when `$SkipPathUpdate` is set.

---
_Generated by [Claude Code](https://claude.ai/code)_

install.ps1
```diff
@@ -1594,20 +1619,286 @@ function Request-OptionalInstall {
...
+function Test-CurlImpersonatePresent {
+  $pathValue = if ($null -eq $env:Path) { '' } else { $env:Path }
+  foreach ($pathEntry in ($pathValue -split ';')) {
+    $directory = $pathEntry.Trim().Trim('"')
+    if ([string]::IsNullOrWhiteSpace($directory) -or -not (Test-Path -LiteralPath $directory -PathType Container)) {
+      continue
+    }
+    try {
+      $hits = @(Get-ChildItem -LiteralPath $directory -File -ErrorAction SilentlyContinue | Where-Object {
+          $_.Name -match '^curl_chrome\d+.*\.(bat|cmd|exe)$'
```
KitsuneKode: This disagrees with the runtime resolver in both directions:

- It matches `curl_chrome131_android.bat` / `curl_chrome99_android.bat` (both ship in the v2.2.1 Windows archive), which `parseWrapper` in `packages/providers/src/shared/curl-impersonate.ts` deliberately rejects as mobile.
- It is chrome-only, where `FAMILY_RANK` also accepts `firefox`, `ff`, `safari` and `edge`.

No impact against the real archive, which carries plenty of desktop chrome wrappers alongside the android ones. It matters for a PATH directory holding only android wrappers (reported present, runtime falls back to plain curl) or only firefox wrappers (reported missing, so the installer re-downloads over a working install). Reusing the same shape as `WRAPPER_PATTERN` would keep the two definitions from drifting further.

---
_Generated by [Claude Code](https://claude.ai/code)_

install.ps1
```diff
@@ -1594,20 +1619,286 @@ function Request-OptionalInstall {
...
+    }
+    # Move-Item into an existing directory nests the extract folder inside it
+    # (`deps\curl-impersonate\extract`). Find-CurlImpersonateWrapperDir recurses,
+    # so that layout would still report success and a later run would treat the
+    # nest as `$existing`. Fail closed if a locked file kept $destDir around.
+    if (Test-Path -LiteralPath $destDir) {
+      throw "Could not replace the existing curl-impersonate directory at $destDir (a file there is in use)."
+    }
+    New-Item -ItemType Directory -Force -Path (Split-Path $destDir) | Out-Null
+    Move-Item -Force -Path $extractDir -Destination $destDir
```
KitsuneKode, line 1927: `$extractDir` lives under `$CacheDir` and `$destDir` under `$DataDir`. The defaults share `%LOCALAPPDATA%`, but a `KUNAI_CACHE_DIR` pointed at another volume makes this directory move fail. It is inside the `try`, so it degrades to the warn plus releases-URL path rather than breaking the install — worth a line of comment, or staging under `$DataDir` so the move is always same-volume.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/cli/test/integration/install-scripts-pwsh.test.ts
```diff
@@ -2031,6 +2046,493 @@ describePwsh("install.ps1 optional dependency consent", () => {
...
+      "$DownloadProgressMinBytes = if ($env:KUNAI_DOWNLOAD_PROGRESS_MIN_BYTES) { [long]$env:KUNAI_DOWNLOAD_PROGRESS_MIN_BYTES } else { 1048576 }",
+      "$DownloadRetryBaseMs = if ($env:KUNAI_DOWNLOAD_RETRY_BASE_MS) { [int]$env:KUNAI_DOWNLOAD_RETRY_BASE_MS } else { 50 }",
+      "$script:LastDownloadHttpStatus = $null",
+      'function Write-Info($m) { Write-Host "-> $m" }',
+      'function Write-Warn($m) { Write-Host "! $m" }',
+      "function Test-Cmd($name) { [bool](Get-Command $name -ErrorAction SilentlyContinue) }",
+      "function Test-RetryableHttpStatus([int]$Status) { return ($Status -eq 408 -or $Status -eq 429 -or $Status -ge 500) }",
+      "function Get-WindowsArch { return 'x64' }",
+      'function Add-UserPath([string]$Dir) { Write-Info "Skipping persistent User PATH update for $Dir." }',
+      "function Register-HelperPath([string]$Dir) {",
```
KitsuneKode: `Register-HelperPath` is the one new function the suite reimplements rather than extracting. Every other helper comes through `extractPs1Function` against the real `install.ps1`, so drift there fails a test; drift here does not — this copy would keep passing while the shipped function changed underneath it.

Adding `"Register-HelperPath"` to the `functions` list above and dropping these six lines would close that, and the stub `Add-UserPath` already defined here gives it the same no-op behavior it has under `-SkipPathUpdate`.

---
_Generated by [Claude Code](https://claude.ai/code)_

KitsuneKode (commented):
Reviewed against the branch checked out locally. All 26 checks green; `bun run verify:doc-paths` passes here. `pwsh` is not available in my environment, so the new installer tests skip locally — the Windows CLI parity job covers them.

Good PR overall: the portable-helper design is the right call, the pin is genuinely correct, and the failure-path coverage is unusually thorough for installer code. One real regression, plus small things — all inline.

## Inline findings

1. **Blocking** — suppressing the curl HTTP/2 upgrade prompt when curl-impersonate is present breaks the HLS relay on Windows. `hls-relay.ts` and Miruro's HTTP/2 probe both use plain `curl`, not `resolveCurlCandidate`, and the pinned archive ships no `curl.exe`. (`install.ps1:1913`)
2. Dead guard: `Test-Cmd 'mpv'` is always true where it's checked. (`install.ps1:1931`)
3. `Register-HelperPath` is the one new function the tests reimplement instead of extracting, so drift in it is invisible. (`install-scripts-pwsh.test.ts:2164`)
4. Duplicate session PATH entry on first install. (`install.ps1:1629`)
5. `Test-CurlImpersonatePresent` accepts android wrappers the runtime rejects, and misses non-chrome families the runtime accepts. (`install.ps1:1688`)
6. `Move-Item` from `$CacheDir` to `$DataDir` is cross-volume if `KUNAI_CACHE_DIR` is redirected. Degrades gracefully. (`install.ps1:1841`)

## Two things worth a decision rather than a change

**Consent posture.** The doc heading went from "Optional dependencies are never installed for you" to "Optional dependencies", and Windows now fetches two third-party executables and writes two permanent User PATH entries with no prompt — opt-out only, and `irm … | iex` users can only opt out through `KUNAI_SKIP_DEPS=1`. Deliberate and documented; flagging it because the repo had been explicit about the opposite rule and the heading change is where that turns over.

**Lifecycle.** Nothing updates the portable copies, and yt-dlp goes stale quickly — a pinned-once `yt-dlp.exe` stops working against YouTube on its own schedule, with no `yt-dlp -U` or re-check anywhere in Kunai. Separately, non-purge `kunai uninstall` leaves `deps\` in place, and neither uninstall path removes the PATH entries. That last part matches the existing BinDir behavior so it is not a new class of problem, just three entries now instead of one. A roadmap row would cover both.

## What checks out

- Pinned digest `f7faa8c4…` matches the real v2.2.1 archive byte-for-byte, and the URL shape matches `setup-curl-impersonate/action.yml` exactly. The lockstep comment on the action input is a good touch.
- The archive extracts flat (`./curl_chrome*.bat` at the root), which is exactly what the fixture reproduces via `tar -czf … -C payloadDir .` — the fixture is not lying about the layout.
- `Get-HelperChecksumEntry`'s looser GNU-style parser with the "exactly one entry or throw" guard is right, and the star and two-space cases are both tested.
- `Invoke-BoundedDownload` uses a bare `return`, so neither `Install-Portable*` leaks stray pipeline output into its `[bool]` return — the trap this shape usually falls into.
- Failure paths are genuinely covered: checksum mismatch leaves no dest, leftover dest is replaced, locked dest fails closed instead of nesting `extract`, already-present skips the network entirely (asserted via `evidence.requests`, not just the log line).
- Realpath-comparing the wrapper directory instead of string-matching is the correct fix for `RUNNER~1` and `/private/var`.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #334, opened 2026-09-04
fix(mobile): release the host input handle so the proof can exit

KitsuneKode:
Stacked on #295 — this targets `feat/mobile-app`, not `main`, so it lands into that PR's branch. It carries the blocking fix and the small edge cases from my review of #295.

## What changed

The Android host proof could not exit: the terminal port opened a stdin reader at construction and never released it, so every path except Ctrl+C printed its output and then hung — including `--help`, the first command the device qualification procedure has a tester run.

### The hang

`createBunTerminalPort` called `Bun.stdin.stream().getReader()` eagerly, and `exitMobile` only sets `process.exitCode`. A live reader holds the Bun event loop open, so the host never drains. Verified against the pre-fix source with stdin as an open pipe:

| Path | Before | After |
|---|---|---|
| `--help` | hangs (SIGKILL at 8s) | exits 0 |
| `--version` | hangs | exits 0 |
| invalid args | hangs | exits 2 |
| host-proof, answer `0` | hangs | exits 0 |
| host-proof, Ctrl+C | exits 0 | exits 0 |

Ctrl+C was the only terminating path, because its interrupt handler cancels the reader — and that is the one cancellation the qualification doc walks through.

Fix: `MobileTerminalPort` declares `close()`, `apps/mobile/src/entry.ts` owns the lifecycle and calls it on every path out (including composition failure, which previously escaped as an unhandled exception because `createMobileEnvironment()` ran while building a call argument), and Android opens stdin on first read so `--help` and `--version` never touch the handle.

### Edge cases in the same surface

- **Trim choice answers.** a-Shell's `read -r` preserves soft-keyboard padding, so `"1 "` was rejected and left the tester in an unbounded retry loop.
- **One cancel entry.** `formatMobileChoiceOptions` already appends `0. Cancel`, so the explicit cancel choice rendered `2. Cancel` beside it.
- **a-Shell output.** `console.log` doubled every newline; the prompt is now written without a trailing space, since nothing on that host can print a partial line.
- **Parser diagnostics.** The reason is surfaced under the existing generic line instead of discarded, and the unknown-option message no longer echoes the rejected token — after a mistyped flag that position holds the URL.
- **HTTP port.** Cancel the response body on the declared-length rejection, matching the streaming branch beside it.

Docs updated in the same change set: `.docs/mobile-terminal-runtime.md` states why `close()` exists, and both device procedures gained an explicit "returns to the prompt on its own" check — the old procedure would have passed a binary that hangs.

Deliberately **not** included, raised on #295 instead: the `.previous`/corrupt-state recovery question (the doc currently specifies fail-closed, so it is a design call), and the eager `Bun.which` ladder in the CLI launcher (changes a `@kunai/core` signature).

## Verification

`test/integration/android-entry-lifecycle.test.ts` bundles the real entrypoint and runs it with stdin held open. All three cases **fail with SIGKILL against the pre-fix source and pass after** — checked both ways round rather than trusting a green run.

**CI on Bun 1.4.0 is green and covers everything**, including the parts I could not run locally. The `Test` job shows `@kunai/mobile:build` executing on a cache miss (`[mobile-build] wrote 3 targets for Kunai 0.3.0`) and `@kunai/mobile:test` as `cache bypass, force executing` — **102 pass, 0 fail across 22 files**, with `build-artifacts.test.ts` (both Android ELF checksums, modes and machine types) and `ashell-launcher.test.ts` passing against freshly built artifacts.

Run locally beforehand: `test:unit` 82 pass · `typecheck`, `lint`, `fmt` · `turbo run typecheck --force` 15/15 · `turbo run lint --force` 13/13 · CLI `boundary-imports` 14 pass · `packages/core` 108 pass · `verify:doc-paths`, `verify:doc-frontmatter`, `verify:doc-coverage` · iOS IIFE rebuilt standalone and exercised under a fake `jsc`, `findForbiddenIosOutputTokens` returns `[]`.

*(Earlier revision of this description asked the reviewer to re-run the Bun 1.4 gate — my container had Bun 1.3.11, which rejects `bun-linux-arm64-android`. CI has now done that; no manual re-run is needed.)*

## Checklist

- [x] Changeset N/A — private preview app, no published release surface
- [x] Guard N/A — no guarded release/version files changed
- [x] `bun run fmt`, `lint`, `test`, `typecheck` pass
- [x] `bun run build` passes — `@kunai/mobile:build` ran on CI, three targets emitted
- [x] Live provider and Discord smokes not applicable; physical-device rows remain pending on #295

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_019T137agG3GDipwShQdTJU3


KitsuneKode:
The still-valid lifecycle, cancellation, parsing, cleanup, and test findings from this PR were reproduced and ported into the current Node/JavaScriptCore mobile architecture in #295. The obsolete Bun-native Android implementation was intentionally not cherry-picked. PR #295 is now rebased directly onto main and is the canonical mobile host-proof change.

PR #336, opened 2026-09-04
fix(docs): give Kanna's dismissal a way back, and stop her taking clicks

KitsuneKode:
Kanna was missing from the deployed site in Zen (a Firefox fork) while still walking in Chrome. Not a rendering bug: `kunai.roamer.dismissed` was set to `"1"` in that profile, and nothing in the app ever removed it.

Two defects of the same shape — something that intercepts input it never showed you, and a door that only opens one way.

## The invisible ×

`.kunai-roamer__close` was `opacity: 0` with live pointer events: an ~18px invisible target at `top:-2px; right:-2px`, sitting over her ear — on the very silhouette you are invited to click to make her talk. A slightly high click retired her permanently, in that browser only, with no confirmation and nothing visible to have aimed at or avoided.

It is now `pointer-events: none`, restored to `auto` in the same rule that makes it visible on hover.

## The way back

Dismissal is now reversible three ways, all through one new module (`lib/roamer-preference.ts`) rather than three components each poking at `localStorage` their own way:

- **Undo** — a pill offered for 12s straight after the click.
- **`?kanna=on`** on any page, with `?kanna=off` as the symmetric way out. Both persist, and the parameter is removed from the address bar once applied so it cannot ride into a copied link or into the URL web analytics records.
- **A live toggle on `/docs/users/kanna`** — the durable, discoverable one. It reports the current state and flips it, and she reacts on the page behind it with no reload.

## Two more found while reading

- **Stale one-shot gate.** `pointer: fine` and `prefers-reduced-motion` were read once at mount. Turning reduced motion on mid-visit left the stylesheet hiding her while the component kept running a `requestAnimationFrame` loop against an invisible fox for the rest of the session; plugging in a mouse later never brought her at all. The gate is now subscribed.
- **She ate clicks meant for the page.** `SETTLE_PX` is 70 and she is 58 across, so at rest her edge is ~41px from your cursor: reaching for anything near where you stopped put the pointer on her, and her click handler took it. Her own doc comment claimed she never covers anything, which was true of the host element and not of the fox. She now stands down — she and her dismiss button both leave hit testing — whenever something interactive is behind her. Over inert page she is still pokeable.

She cannot solve that one by moving, which is why the fix is not a dodge: a pointer that has drifted onto a resting fox has travelled far less than `NOTICE_PX`, so she never notices and never walks away. There is a test for exactly that. Dwell timing was the obvious alternative and is wrong — aiming at a link is itself a pause.

## Scope

I swept the docs app for both classes. Each had exactly one instance, both in the roamer: every other `opacity: 0` is a `@keyframes` step or an SVG crossfade layer, and the roamer was the *only* web-storage user in the entire app. No cookies; Vercel Analytics is cookieless.

## Verification

The CSS guard is written against the **class**, not the selector: any future rule under `.kunai-roamer` that sets `opacity: 0` without `pointer-events: none` fails `test/roamer-affordance.test.tsx`, and it carries a self-check proving the checker is not vacuous.

Beyond unit tests, both fixes were driven end to end in real headless Chrome over CDP, each with a **negative control** that re-injects the old declaration and confirms the bug reappears — without that, a green check proves nothing:

- 16/16 on dismissal, restore, the URL parameter, and the docs toggle, including that client-side routing still works after the history rewrite and that back does not land on the spent parameter.
- 4/4 on yielding, where the probe *sweeps pointer positions until she genuinely parks on a sidebar link* — no rigged transform — then checks who wins `elementFromPoint`.

Gates, all re-run with `--force` past the turbo cache: root `typecheck`, `test` (255 pass / 25 skip / 0 fail on the CLI integration suite), `lint`, `fmt:check`, `verify:doc-paths`, `verify:doc-coverage`, `lint:anti-slop:changed`, plus a clean `next build` from a deleted `.next`.

Two anti-slop findings remain in `kunai-fox-roamer.tsx`; both are on lines that exist unchanged on `main` and are untouched here.

## Note for existing visitors

Anyone who already dismissed her stays dismissed — that is correct, it was their choice. They get the restore routes above. My own Zen profile is one of them.

## Summary by CodeRabbit

- **New Features**
  - Added a documentation-site toggle for enabling or dismissing the roaming fox.
  - Added a 12-second undo option after dismissal.
  - Added URL controls to enable or disable the fox with `?kanna=on` or `?kanna=off`.
- **Bug Fixes**
  - The fox now yields interaction when positioned over links and buttons.
  - Preferences stay synchronized across browser tabs and documentation pages.
  - The fox responds to pointer and reduced-motion setting changes.
- **Documentation**
  - Documented the fox’s controls, availability, preferences, URL options, and accessibility behavior.

### DIFF
--- a/apps/docs/app/styles/fox.css
+++ b/apps/docs/app/styles/fox.css
@@ -257,7 +257,8 @@ a:focus-visible .kunai-nav-title .kunai-fox {
    Kanna loose on the page. Fixed and transformed every frame from a ref, so
    moving her never queues a React render. Pointer events pass straight through
    the host — only she, and her dismiss button while it is actually visible, are
-   clickable, because a companion that eats clicks is a companion people close. */
+   clickable, and both stand down while she is in front of something the page
+   wanted clicked, because a companion that eats clicks is one people close. */
 
 .kunai-roamer {
   position: fixed;
@@ -380,6 +381,19 @@ a:focus-visible .kunai-nav-title .kunai-fox {
 .kunai-roamer:hover .kunai-roamer__close {
   opacity: 1;
   pointer-events: auto;
+}
+
+/* She stands down while she is between the reader and something clickable —
+   see `updateYield` in the component. Both controls go at once: a dismiss
+   button still reachable through a fox that is pretending not to be there
+   would be the invisible target again, wearing a different hat.
+
+   Last on purpose. It ties on specificity with the `:hover` rule above, so
+   order is what decides, and the rule that gives a click back to the page has
+   to be the one that wins. */
+.kunai-roamer[data-yield="true"] .kunai-roamer__fox,
+.kunai-roamer[data-yield="true"] .kunai-roamer__close {
+  pointer-events: none;
 }
 
 /* ── Undo ───────────────────────────────────────────────────────────────────
--- a/apps/docs/app/styles/fox.css
+++ b/apps/docs/app/styles/fox.css
@@ -256,8 +256,8 @@ a:focus-visible .kunai-nav-title .kunai-fox {
 /* ── Roamer ─────────────────────────────────────────────────────────────────
    Kanna loose on the page. Fixed and transformed every frame from a ref, so
    moving her never queues a React render. Pointer events pass straight through
-   the host — only she and her dismiss button are clickable, because a companion
-   that eats clicks is a companion people close. */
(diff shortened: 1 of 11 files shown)


KitsuneKode:
@coderabbitai review

KitsuneKode:
@coderabbitai review

PR #340, opened 2026-09-08
fix(test): honest provider-matrix rows plus headless mpv decode check

KitsuneKode:
Matrix was misreporting two providers: youtube false-red (NDJSON check lines broke the single-JSON parse) and anidb search-stage failures lost to stderr. Plus AniDB parsed 503 outage pages as zero results.

- matrix-harness.mjs (shared, unit-tested): NDJSON primary-payload parse, stdout->stderr fallback, 5xx/captcha/WAF environment-network rules
- anidb searchAnidb passes reportStatus:true so HTTP errors throw AnidbHttpStatusError; smoke reports structured JSON
- mpv-playback.smoke.ts + test:live:mpv: headless decode proof with production header split, redacted logs
- Live evidence: matrix 2 healthy (vidlink, youtube, both mpv-decoded), 0 harness-failure; videasy/rivestream CDN blocks, anidb site 503, miruro WAF, allanime zero-links are upstream, mpv-proven

## Summary by CodeRabbit

* **Bug Fixes**
  * AniDB search now reports service errors, such as maintenance or access-block pages, instead of incorrectly showing no results.
  * Rivestream verifies stream candidates before selection and automatically tries another mirror when playback is definitively unavailable.
  * Improved Rivestream proxy playback by forwarding the correct stream headers.
  * Fixed Videasy playback on supported CDN hosts by omitting an incompatible request header.

* **Testing**
  * Added optional headless mpv playback checks to validate actual stream decoding.

PR #337, opened 2026-09-04
fix(providers): harden stream resolution, CDN reachability failover, and mpv handoff

KitsuneKode:
## Summary

This PR addresses stream reachability gaps, upstream schema/maintenance drift, and mpv handoff edge cases across production providers (`videasy`, `vidlink`, `rivestream`, `anidb`, `allanime`, `miruro`, `youtube`).

### Key Hardening & Fixes

1. **Miruro & Rivestream Reachability Failover**:
   - Integrated `probeStreamReachability` inside candidate resolution within `runProviderCycle`.
   - Prevents dead, 429 rate-limited, or expired CDN endpoints (e.g. `pewe` CDN returning HTTP 429 on Miruro, or ByteDance ImageX 1004 on Rivestream) from being returned as false successes. The engine cycles cleanly to working mirrors (e.g. `moo` on Miruro).
   - Preserved `ProviderCycleFailureError` across error boundaries.

2. **Rivestream DASH Sniffing & CORS Origin Header**:
   - Fixed stream candidate protocol and container inference: `.mpd` URLs are now properly mapped to `protocol: "dash"` and `container: "mpd"` instead of defaulting to MP4, giving mpv the correct demuxer hints.
   - Added `origin: "https://www.rivestream.app"` alongside `referer` for CDNs requiring CORS Origin verification.

3. **AniDB Maintenance Detection**:
   - Detected upstream HTTP 503 and HTML maintenance pages (`<title>Under Maintenance</title>`) in `anidbFetchText()`.
   - In `anidbProviderModule.search()`, catch 503 maintenance errors and return `null` so `ProviderRegistry` marks the provider offline instead of misleadingly displaying 0 results.

4. **AllAnime Persisted Query Drift**:
   - Added `AllMangaQueryDriftError` and `isAllMangaPersistedQueryNotFound()` to detect `PersistedQueryNotFound` when upstream rotates GraphQL query hashes.
   - Classifies as non-retryable `code: "parse-failed"` with actionable diagnostics instead of quietly returning empty sources.

5. **YouTube yt-dlp & mpv Handoff Hardening**:
   - Added `/ba` fallback to `defaultYtdlPlaybackFormat()` (`"bv*+ba/b/ba"`) and `DEFAULT_MPV_YTDL_FORMAT`. This prevents mpv from failing on audio-only uploads, music tracks, and podcasts that lack a video track (`bv*`).
[...]

PR #345, opened 2026-09-08
fix(providers): one resolve gate, and the four providers that were reporting success for streams that cannot play

KitsuneKode:
Every defect here is the same shape: **a provider reports success for something that cannot play.** Resolve-time success is not the contract users care about; decode is.

The matrix scored 2/7 when this started. It now scores 4 healthy with 0 harness failures, and each of the three reds names a real upstream condition instead of a silent zero.

## What users get

- **Videasy** — the default movie/series provider — shipped a stream the CDN refuses, **in the released build**. It carries two origins that are not interchangeable: the front-end we impersonate when calling its API (`cineby.at`) and the origin the media CDN's hotlink rule accepts. The resolve path reused the first as the second. Measured against `moon.peakstorm.top` with the cineby watch page as Referer: `vidking.net` → 200, `cineby.at` → 403, absent → 403.
- **Rivestream** reported `provider:success` for a playlist whose 349 segments answered `{"code":1004,"error":"domain forbidden"}`. Cycling stopped there and never reached `citadel`, which plays.
- **AllManga** returned zero streams for every title because mkissa rotated its client crypto and **every** pinned constant moved at once — build id, all four mask fragments, the derivation constants, the boot prefix and separator, the boot payload *field order*, and the persisted-query hash.
- **AniDB** answered a site-wide `503` maintenance page, which was scraped into "no results" for every query alike — an outage wearing the costume of "no such anime".

## The systemic fix

Videasy and Rivestream were the same bug wearing different clothes: a gate that assembled its probe separately from the thing it shipped. `verifyCandidateStream` is now the one resolve gate and it takes **the candidate**, not a URL plus separately-built headers, so the verified request and the played request cannot drift apart. A coverage test enforces that every production provider uses it or carries a documented runtime reason not to.

Consolidating surfaced three more defects:

- `direct-stream-source` set `streamReachabilityVerified` from a **timed-out** probe, promoting an unproven stream to "provider-attested" and switching off the playback preflight that would have caught it.
- Gate budgets could not reach a verdict. An HLS gate is three sequential round trips, measured at 2.1–2.9s; the shared budget was 3s and Videasy carried its own tighter 2.5s copy. A cut-short probe reports `timeout`, which is deliberately read as "not proven dead" — so an under-budgeted gate does not fail loudly, it passes dead streams *intermittently*. That is the "sometimes it works" behaviour.
- AllManga had no gate and a candidate timeout too small to contain one.

## Verification

Decode, not probes. `bun run test:live:mpv` spawns headless mpv with the production header split and counts frames.

| Provider | Result |
| --- | --- |
| videasy | 3/3 decoded, ~2.0s |
| vidlink | 3/3 decoded, 1.1–1.7s |
| youtube | 3/3 decoded, 2.5–3.5s |
| rivestream | 3/3 decoded, 3.2–5.5s |

Twelve consecutive runs, same host every time. AllManga was proven end to end once (4 candidates from `video.wixstatic.com`, mpv decoding h264 1080p `alang=jpn`) before its network started answering `NEED_CAPTCHA`.

`KUNAI_MATRIX_PLAYBACK=1 bun run test:live:matrix` now carries that decode check, and the report states which kind of evidence it holds so a run can never be mistaken for the stronger claim.

## Not done, deliberately

- **Miruro** stays exempt from the gate. Its per-candidate budget is 4.8–5s after clamping and must also cover its pipe call, so a gate would be cut short and pass everything — coverage in name only. Separately, all four of its mirrors answer Cloudflare's managed challenge; twelve impersonation profiles were tried and all twelve are challenged, so no newer curl-impersonate build fixes it. The relay path *is* verified wired (four `POST /rpc/miruro` calls with correct upstream URLs), it just needs an egress the WAF tolerates.
- **AllManga's and AniDB's gates have not met a live source** — both providers are blocked upstream right now. They are unit-tested and shared with four mpv-verified providers, but unconfirmed is not confirmed.

Residue is tracked in `.plans/provider-playback-resilience.md`.

## Relationship to #340

This overlaps #340 substantially and the two cannot both merge as-is. Five of the six CodeRabbit findings on #340 applied here too — the mpv verifier was adopted from that branch, and the parser and gate had grown the same shapes — and they are fixed in this branch.

Two points where this branch disagrees with #340, both backed by measurement:

- #340 fixes Videasy by **omitting** `Origin`, on the reading that the CDN 403s any request carrying one. With the cineby referer I measured no-origin → **403** and `vidking.net` → **200**. Note the gate is intermittent: re-measured later the same day, all nine referer×origin combinations returned 200. Sending the origin a real vidking.net player would send works under both observed regimes; omitting it does not.
- #340 sets its Rivestream gate budget to 3s, which is the value proven here to flip its verdict between consecutive runs.

#340's flowcast per-source proxy headers are genuinely additive and not duplicated here.

## Summary by CodeRabbit

- **Bug Fixes**
  - Restored AllManga playback following an upstream client update.
  - AniDB outages now report service errors instead of appearing as empty search results.
  - Providers now verify streams before accepting them, improving fallback to working mirrors.
  - Rivestream skips mirrors whose media segments cannot be played.
  - Videasy playback now avoids CDN authorization failures caused by mismatched origins.

- **Tests**
  - Added live checks for provider playback, stream verification, and AllManga crypto freshness.
  - Improved provider health reports to distinguish upstream, network, and test-harness failures.


KitsuneKode:
## Principal Code Review & Technical Audit Findings

Conducted a thorough audit and code review of this PR across architecture, standards, and runtime behavior, comparing it against #340 and #337.

### Strengths & Architectural Achievements
1. **Single Resolve Gate (`verifyCandidateStream`)**: Direct evaluation of candidate stream objects (`url`, `headers`, `protocol`, `referrer`) completely eliminates probe/play drift.
2. **Empirical Gate Sizing**: Raising the default gate timeout to 6.0s correctly accommodates measured 2.1–2.9s 3-hop HLS handshakes.
3. **Build 166 Crypto Parity**: Flawless reverse engineering of mkissa's `Rf` config (constants, field order, and query hash) with golden vector unit tests and live freshness monitoring.
4. **Headless MPV Verification Harness**: Spawning `mpv` with production header splitting (`--vo=null --ao=null --frames=30`) verifies true decode capability rather than superficial manifest 200s.

---

### Critical Bugs & Risks to Address Before Merge

#### 1. [Blocking] Error Swallowing Drops Rivestream Dead Mirror Quarantine
In `packages/providers/src/rivestream/direct.ts:434-458`:
When `resolveRivestreamProviderCandidate` throws a `ProviderCycleFailureError` with `endpointScoped: true` and `failureClass: "candidate-blocked"`, the outer `try/catch` catches it, wraps it in a new `ProviderHttpError`, and throws a new `ProviderCycleFailureError` with `failureClass: "candidate-empty"`, **completely dropping `endpointScoped: true`**.
- In `packages/core/src/provider-cycle-engine.ts`, `classifyEndpointFailureFromCycleFailure` requires `failure.failureClass === "candidate-blocked" && failure.endpointScoped === true`.
- **Impact**: `endpointHealth.recordFailure(...)` is never invoked, meaning dead mirrors are never quarantined and will be re-walked every play (breaking commit `09278cb9`).
- **Remedy**: Add an early re-throw at the top of the catch block:
  ```ts
  if (error && typeof error === "object" && "name" in error && error.name === "ProviderCycleFailureError") {
    throw error;
  }
  ```

#### 2. [Blocking] Unhandled Exception on Zero Stream Candidates in Rivestream
In `packages/providers/src/rivestream/direct.ts:958-984`:
If an upstream mirror returns 0 usable streams, the function returns `{ provider, sourceId, streams: [], ... }` instead of throwing. `runProviderCycle` treats this as a successful resolution, passing `selectableStreams = []` to `selectReadyStream`, which throws an unhandled `Error("No ready stream candidates")` that halts the entire provider cycle.
- **Remedy**: If `streams.length === 0`, throw `createProviderCycleFailureError(candidate, { failureClass: "candidate-empty", ... })`.

#### 3. [High Risk] Budget Ceiling Inversion in `fast` Startup Mode
- In `packages/core/src/provider-attempt-budget.ts`, the `fast` mode attempt budget is 6,000ms, making the candidate timeout ceiling 4,800ms (`6,000 * 0.8`).
- If the resolve-gate timeout is 6,000ms, the candidate timeout controller aborts at 4.8s before the gate probe can complete.
- In AllManga (`packages/providers/src/allmanga/direct.ts:647`), `candidateTimeoutMs: 9_000` is hardcoded without clamping, exceeding the entire 6,000ms attempt budget in `fast` mode.
- **Remedy**: Clamp gate timeout to `Math.min(resolveGateTimeoutMs, candidateTimeoutMs - 500)` and clamp AllManga's candidate timeout via `providerCycleCandidateTimeoutMs`.

#### 4. [High Risk] AllManga Signal Scoping
In `packages/providers/src/allmanga/direct.ts:678`, `verifyCandidateStream` receives `signal: context.signal` (the provider attempt signal) rather than `cycleContext.signal` (the candidate-scoped signal). If a candidate times out, its background probe request will not be aborted immediately.
- **Remedy**: Pass `signal: cycleContext.signal`.

#### 5. [Collision Warning] Merge Conflict with PR #326 in `anidb/client.ts`
PR #326 introduces `RELAY_HOP_HEADER = "X-Kunai-Relay-Hop"` to distinguish stale relay 404s (`unknown-provider`) from genuine upstream 404s. PR #345's `isFingerprintRetryableStatus` refactor must preserve this check when rebasing, otherwise stale relays will throw `AnidbHttpStatusError(404)` and resurrect the anime lane outage.

---

### Recommended Cherry-Picks from #340 & #337
Before closing #340 and #337 as superseded, consider cherry-picking:
1. **From #340**: Flowcast per-source proxy headers (`parseRivestreamProxyHeaders` and `resolveRivestreamStreamHeaders`).
2. **From #337**:
   - YouTube audio-only fallback: `DEFAULT_MPV_YTDL_FORMAT = "bv*+ba/b/ba"`.
   - YouTube explicit `--ytdl=yes` in mpv launch arguments.
   - YouTube rental/purchase terminal failure classifier.
   - Rivestream DASH container inference (`.mpd` $\to$ `protocol: "dash"`).

PR #346, opened 2026-09-08
fix(providers): unified provider playback resilience, decode gates, and headless mpv verification

KitsuneKode:
## Summary
Consolidates, reconciles, and proves the fixes across **PR #345**, **PR #340**, **PR #337**, **PR #329**, and **PR #326** into a unified, clean, and comprehensively verified branch.

### Key Improvements Unified
1. **Decode-Time Verification & Resolve Gate (PR #345)**:
   - Implemented `verifyCandidateStream` with segment probing and budget isolation (6s) so dead master playlists (e.g. ByteDance ImageX 1004 on Rivestream) fail before reaching the player.
   - Videasy media origin set to `VIDKING_ORIGIN` (`https://www.vidking.net`) matching the CDN hotlink rule.
   - AllManga `build 166` crypto with fresh derivation constants, mask fragments, and GraphQL hashes.
   - Actionable `NEED_CAPTCHA` diagnostics instead of obscure parsing crashes.
2. **Headless MPV Playback Smoke Test (PR #340)**:
   - Added `mpv-playback-verify.ts` and `mpv-playback.smoke.ts` running headless MPV (`--frames=30 --vo=null --ao=null`) to prove actual video frame decoding.
   - Unified Flowcast per-source proxy headers and DASH `.mpd` MIME detection into Rivestream direct resolution.
3. **YouTube Playback Hardening (PR #337)**:
   - Default format selector widened to `"bv*+ba/b/ba"` to support audio-only, music, and podcasts without crashing.
   - Enforced `--ytdl=yes` in MPV args to prevent user `mpv.conf` overrides (`ytdl=no`) from breaking YouTube playback.
   - Added purchase/rental failure classification in metadata failure classifier.
4. **Relay & Anime Hardening (PR #326 & PR #329)**:
   - Forwarded `x-playback-environment: webkit` on the relay to prevent BunnyCDN HTTP 429 rate-limiting on VidLink.
   - Terminated curl argument parsing (`--`) in AniDB and Miruro to prevent flag injection.
   - AniList search outage fallback to sibling anime providers.
   - Cross-catalog identity lifting in stream request adapter.
   - Resolved merge conflict in `packages/providers/src/anidb/client.ts` cleanly.

---

#### Test Suite
- **Full Typecheck**: 14/14 packages passed, 0 errors (`bun run typecheck --force`).
- **Lint**: 12/12 packages passed, 0 errors, 0 warnings (`bun run lint`).
[...]

PR #347, opened 2026-09-08
fix(providers): one resolve gate, so a provider cannot attest a stream it never proved

KitsuneKode:
## Why

Two providers were reporting `provider:success` for streams that cannot play, and neither was visible from the CLI — resolve looked clean and mpv failed.

- **Rivestream** accepted a master playlist whose segments answered `domain forbidden`. Cycling stopped at that server instead of walking on to one that plays.
- **Videasy** probed with `Origin: vidking.net` and shipped with `Origin: cineby.at`. The CDN accepts the first and refuses the second, so its check attested a request shape production never made. **This one is in the released 0.3.0 build.**

## What changed

`verifyCandidateStream` takes the **candidate**, not a URL plus separately-assembled headers — that is the whole point, because it makes the probed request and the shipped request the same object. `selectVerifiedStream` walks a candidate's rungs, probes one per host, and reports which hosts refused.

Videasy, VidLink, Rivestream, AllManga and AniDB resolve through it. YouTube and Miruro are exempt for reasons about the runtime, recorded in `provider-resolve-gate-coverage.test.ts`.

Three things this had to get right:

| | Problem | Fix |
|---|---|---|
| **Budget** | The attempt budget clamps a candidate to 80% of its timeout — 4.8s on `fast`. A 6s gate cannot be scheduled inside it, so it timed out and passed everything. | `resolveGateBudgetMs` sizes each probe against the *clamped* figure, allowing for the retry probe. |
| **Failure class** | A refused stream was rebuilt as a generic `ProviderHttpError`, which relabelled it `candidate-empty` and dropped the endpoint flag. Rivestream re-walked all eleven services on every play and learned nothing. | The gate's verdict is re-thrown intact: `candidate-blocked` with `endpointScoped: true`, so endpoint health can quarantine the mirror. |
| **Inventory** | A service whose 1080p rung is dead and whose 720p rung plays passed on the 720p probe, then shipped the whole inventory — including the rung the gate had just refused, which is the one startup selection prefers. | `dropRefusedStreams` prunes them. |

## Evidence

Live headless-mpv decode (`--frames=30`), all resolving and decoding:

| Provider | Latency |
|---|---|
| videasy | ~2.0s |
| vidlink | 1.1–1.7s |
| youtube | 2.5–3.5s |
| rivestream | 3.2–5.5s |

`bun run typecheck --force`, `bun run lint --force`, `bun run test --force` (23/23, 0 cached) green on this commit alone.

## Review notes

- Gate exemptions need a **runtime** reason, and the test asserts one exists.
- `resolveGateBudgetMs` divides by `RESOLVE_GATE_MAX_PROBES` because `runStreamHealthCheck` retries once when the first attempt reaches no verdict — the candidate has to afford both.
- Fixture CDNs in `allmanga.test.ts`, `anidb.test.ts` and `providers.test.ts` now answer the segment hops; without that every resolve fixture fails as a dead mirror, which is what the gate is there to reject.

https://claude.ai/code/session_01JwGMDjtMd6ozHYGXaCr13N

---

## The stack

Merge bottom-up. Each was gated independently with `--force` (no turbo cache replays).

| | PR | Δ | |
|---|---|---|---|
| 1 | **#347** ← you are here | +1400/−74 | the gate, and every provider that adopts it |
| 2 | #348 | +558/−65 | AllManga build 166 + rotation detection |
| 3 | #349 | +698/−45 | AniDB outage honesty, stale-relay 404, curl `--` |
| 4 | #350 | +207/−33 | VidLink header, `--ytdl=yes`, audio-only selector |
| 5 | #351 | +1217/−331 | headless-mpv decode proof, honest matrix |

This replaces **#346**, which carries the same content as one 75-file diff. #345, #340, #337 and #329 are subsumed — except one deliberate omission from #337, noted in #348.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved playback reliability by verifying streams before selecting a provider or server.
  - Automatically skips unavailable mirrors and continues trying healthy alternatives.
  - Remembers failed mirrors to avoid repeatedly selecting sources that cannot play.
  - Fixed certain Videasy streams that could resolve successfully but fail with a 403 error during playback.
  - Improved handling of stream headers and media origins for supported providers.
- **Tests**
  - Added coverage for stream verification, fallback behavior, mirror health, and provider playback scenarios.

### DIFF
--- a/.changeset/rivestream-verified-source.md
+++ b/.changeset/rivestream-verified-source.md
@@ -1,0 +1,14 @@
+---
+"@kitsunekode/kunai": patch
+---
+
+Stop Rivestream selecting a source that cannot play.
+
+Rivestream reported success for a playlist whose segments were refused by a
+third-party CDN, so cycling stopped at the first server that merely responded
+and never reached one that plays. It now proves a source before accepting it,
+walking that source's qualities and moving to the next server only when every
+distinct host has refused.
+
+A refusal is also recorded against that server, so a mirror proven dead is
+skipped on later plays instead of being re-walked every time.
--- a/.changeset/videasy-media-origin.md
+++ b/.changeset/videasy-media-origin.md
@@ -1,0 +1,15 @@
+---
+"@kitsunekode/kunai": patch
+---
+
+Fix Videasy handing the player a stream the CDN refuses.
+
+Videasy carries two origins that are not interchangeable: the front-end we
+impersonate when calling its API (`cineby.at`), and the origin the media CDN's
+hotlink rule accepts. The resolve path reused the first as the second, so the
+default movie/series provider shipped a URL that answered 403 the instant it was
+resolved — while its own resolve gate probed the correct origin and reported
+success, which suppressed falling back to a source that would have played.
+
+The stream origin is now a single constant with no override, so the request that
+is verified and the request that is played cannot drift apart.
--- a/.plans/provider-playback-resilience.md
+++ b/.plans/provider-playback-resilience.md
@@ -1,0 +1,54 @@
+# Provider playback resilience
+
+Residue from the 2026-09-08/09 provider health pass. This tracks what is **not**
+closed; the landed work is the commits themselves.
+
+**Framing:** every defect in this pass was the same shape — _a provider reports
+success for something that cannot play_. Resolve-time success is not the
+contract users care about; decode is. `verifyCandidateStream` is the single seam
+that keeps the probed request and the shipped request identical.
+
+## Closed so far
+
+- **Videasy shipped a dead URL past its own gate.** It verified with
+  `Origin: vidking.net` and shipped `Origin: cineby.at`; the CDN accepts the
+  first and refuses the second. The stream origin is now one constant with no
+  override. This was in the released build.
+- **One resolve gate.** `verifyCandidateStream` takes the candidate, so the two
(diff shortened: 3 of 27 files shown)


KitsuneKode:
The **macOS CLI parity** red here is not this PR's doing, and it is not random.

It is `install.sh lifecycle contract > a failed reclaim with zero poll observes the activation deadline`, failing at **541ms against a 500ms budget** — twice on this commit, including on a re-run. Three things establish it is pre-existing:

- this branch touches no installer file;
- the test landed on `main` in `91cca8ad` (2026-08-25), two weeks before this work;
- **#348 and #349 contain this exact commit** and their macOS job passed.

The budget is measuring the wrong thing: `runInstallShWithin` races the whole `bash install.sh` run — startup, the shimmed `mv`, the fixture download and its checksum — against the deadline, while the activation lock wait it is named for is 40ms of that.

Fixed separately in **#352**. Merge that first, bring `main` into this branch, and the red clears.

https://claude.ai/code/session_01JwGMDjtMd6ozHYGXaCr13N

KitsuneKode:
Rebased onto #352 so this PR now carries the installer-flake fix as its first commit. macOS CLI parity should no longer be the inherited 541ms-vs-500ms failure. Merge order is still bottom-up: **#352 or this PR first** (they share that commit), then #348 → #349 → #350 → #351 → the live-followups PR.

PR #358, opened 2026-09-10
fix(miruro): stop reading a dead upstream server as a Cloudflare block

KitsuneKode:
## What was wrong

Miruro looked dead. It wasn't — it was the only working anime provider, and a predicate was hiding it.

```ts
function isCloudflareHtmlBody(body: string): boolean {
  const head = body.slice(0, 200).toLowerCase();
  return head.includes("<!doctype html") || head.includes("<html") || head.includes("just a moment");
}
```

That is true of Cloudflare's block page. It is equally true of the mirror's own error page:

```
HTTP 444 — <title>502 upstream unreachable</title>
```

So one dead backing server read as a region-wide WAF block, and the misclassification cascaded:

1. the false WAF hit set `wafLikely` for the second mirror, which **suppressed the curl fallback** there
2. two hits crossed `MIRURO_WAF_FAIL_FAST_THRESHOLD` and threw the WAF error
3. `shouldStopAfterFailure` matched `"cloudflare"` in that message and **stopped the whole cycle**

Every server ordered behind the dead one was skipped, and the error told the user to configure a relay — advice that cannot help, because nothing was blocking them.

## Live measurements (2026-09-11)

Per-server `sources` probe, One Piece ep 1159:

| Server | Backing | Result |
| --- | --- | --- |
| `pewe` | AniDB | 444 — anidb.app is serving "Under Maintenance" |
| `moo` | AnimeGG | **200, 4 streams** |
| `bee` | Anikoto | **200, 2 streams** |
| `ally` | AllManga | 444 — api.mkissa.net is 403 |
| `bonk` | — | **200, 7 streams** |
| `kiwi` | owocdn | **200, 4 streams** |
| `hop` | — | 444 |

Transport, same window: Bun `fetch` is **always** CF-403'd (its TLS fingerprint); curl clears the edge on both mirrors given the `sec-fetch-*` triplet the code already sends. The 2026-08-17 manifest note claiming curl also received CF 403 was measured through the broken predicate and is corrected here.

**Live smoke: `ok: false` → `ok: true`**, resolving through `moo` from `www.animegg.org` in 4.6s. The trace now shows `pewe` failing and the cycle *continuing* rather than aborting.

## How detection changed

The new marker set contains only strings that appear on Cloudflare's own interstitials: `attention required`, `cf-error-details`, `cf-wrapper`, `just a moment`, `checking your browser`, `__cf_chl`.

Deliberately **absent**: the word `cloudflare`, and `/cdn-cgi/`. The origin sits behind Cloudflare, so its own error page carries the `cloudflareinsights.com` beacon *and* a `/cdn-cgi/challenge-platform/scripts/jsd` script — matching on either reproduces the exact bug being fixed. Both fixtures in the new tests are trimmed from the live captures, so the regression is pinned to real bodies rather than a guess at their shape.

ani-cli arrives at the same design from the other direction: it tests only for `"Just a moment"` (`ani-cli:171`, v5.0.4), never for HTML-ness.

A 444/502/503/504 that is *not* a Cloudflare page is now reported as `HTTP 444 (upstream server unavailable)` — a per-server fact that does not touch the WAF counter and does not stop the cycle.

## Block message

When a real block does happen, the message now names the cheapest fix that can work. A user on plain curl is told to install curl-impersonate first; a user who *already* impersonated and is still blocked is told the block is region-wide and pointed at the relay. This mirrors ani-cli, which raises `"Blocked by cloudflare. Try installing curl-impersonate"` only while `$curl_exe` is still plain `curl`.

`resolveCurlCandidate()` already returned `{impersonates, profile}` for exactly this purpose; nothing was reading it.

## Also in here

- **`ProviderManifestStatus` drops `"experimental"` and `"research"`.** No code ever branched on them, so both rendered identically to `production` while reading like they gated something. The only three sites setting them are a template, a research-only module and a relay test fixture — none registered. Real research standing already lives in `packages/providers/src/research.ts`, and the real gate is `loadProductionProviderModules()`. `candidate` is now documented as what it actually is: a display suffix in the Tracks panel and provider picker, with no effect on registration, ordering, fallback or health.
- **`.docs/providers.md` listed `allanime` as `anime, series`.** Its manifest is `mediaKinds: ["anime"]` and its resolver has no series path.

## Verification

`typecheck` 14/14 · `test` 23/23, 0 fail · `lint` 0 errors · `fmt` · `verify:doc-paths` all resolve — each run with `--force` to defeat turbo cache replay. Live: `test:live:miruro` green.

## Not in scope

`MIRURO_SERVER_TRY_ORDER` still leads with `pewe` and demotes `kiwi` to 10th as "stalling/444-returning", which today's data contradicts. Left alone on purpose: that ranking is one machine at one moment, and `pewe`/`ally` are only down because AniDB and AllManga are down — both may return. Worth revisiting with data over time rather than a single sample.

## Summary by CodeRabbit

- **Updates**
  - AllAnime is now documented as covering anime content only.
  - Provider status options have been simplified to “production” and “candidate.”
  - The template and Cineby providers now display as candidates.

- **Bug Fixes**
  - Improved Miruro error reporting distinguishes Cloudflare blocks from unavailable upstream servers.
  - Miruro can continue checking remaining servers after temporary 444, 502, 503, or 504 responses.
  - Cloudflare detection is more accurate across supported connection methods.

- **Documentation**
  - Updated Miruro guidance with current connection behavior and clearer troubleshooting details.

### DIFF
--- a/.docs/providers.md
+++ b/.docs/providers.md
@@ -518,7 +518,7 @@ modes.
 | `rivestream` | movie, series | direct-http | `packages/providers/src/rivestream/direct.ts` |
 | `videasy`    | movie, series | direct-http | `packages/providers/src/videasy/direct.ts`    |
 | `anidb`      | anime         | direct-http | `packages/providers/src/anidb/direct.ts`      |
-| `allanime`   | anime, series | direct-http | `packages/providers/src/allmanga/direct.ts`   |
+| `allanime`   | anime         | direct-http | `packages/providers/src/allmanga/direct.ts`   |
 | `miruro`     | anime         | direct-http | `packages/providers/src/miruro/direct.ts`     |
 | `youtube`    | video         | direct-http | `packages/providers/src/youtube/direct.ts`    |
 
--- a/apps/cli/src/domain/types.ts
+++ b/apps/cli/src/domain/types.ts
@@ -258,7 +258,8 @@ export interface ProviderMetadata {
   readonly isYoutubeProvider: boolean;
   readonly providerLane: ProviderLane;
   readonly catalogIdentity?: "provider-native" | "anilist" | "tmdb";
-  readonly status?: "production" | "candidate" | "experimental" | "research";
+  /** Display-only; mirrors `ProviderManifestStatus` in `@kunai/core`. */
+  readonly status?: "production" | "candidate";
   readonly domain?: string;
 }
 
--- a/apps/docs/lib/generated-metadata.json
+++ b/apps/docs/lib/generated-metadata.json
@@ -1,7 +1,7 @@
 {
   "version": "0.3.0",
   "cliVersion": "0.3.0",
-  "cliSourceFingerprint": "9399f4e19c122a598078ee554d5ef0659a4c62985eef9ea2a9a7b0aa8ac691de",
+  "cliSourceFingerprint": "970f2b0ea4a34e8034a0210a63fca94ffc6e1924b36f64a03abe50b0cd33ce29",
   "docsContentFingerprint": "6cf519d315982a3c8c0617891325b79a3179587b3761e6097f012d298ee5dfc7",
   "featureStatusRevision": "5a88d5e2542655bbbfab1e4d050f51632669ac5d46e2b22ef14596f3684a8257",
   "commandCount": 82,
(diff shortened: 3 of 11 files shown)


KitsuneKode:
@coderabbitai review

KitsuneKode:
@coderabbitai review

PR #364, opened 2026-09-11
feat(miruro): use every official mirror, and learn new ones from Miruro's status page

KitsuneKode:
**Stacked on #363** (which is stacked on #358). Merge those first.

## Why

Every Miruro backend is reached through `/api/secure/pipe` on one host, so the mirror list is the anime lane's single point of failure. Kunai shipped two of the four that work.

`.tv` and `.to` were excluded as "TLS-dead". They are not — all four `www.` mirrors served the pipe on 2026-09-11:

```
www.miruro.bz   200  x-obfuscated: 2
www.miruro.ru   200  x-obfuscated: 2
www.miruro.to   200  x-obfuscated: 2
www.miruro.tv   200  x-obfuscated: 2
www.miruro.com  200  (landing page, no pipe)
```

That old reading came from a network whose reachability to them **flaps**: the same host timed out at 12s, failed fast on `-6`, then answered over dual-stack minutes later. Which is also why order now matters as much as the list.

## Three layers

1. **The shipped list**, now all four, so a cold start waits on nothing.
2. **Miruro's own status page.** `status.miruro.com` is Uptime Kuma with public JSON (`/api/status-page/miruro` + `/api/status-page/heartbeat/miruro`). Its `mirrors` group names hosts Kunai doesn't ship with and reports which are down. Read **in the background**, cached 30 min, so no request ever waits on it, and a failed read isn't retried until the TTL expires.
   - Only `miruro.<tld>` names are accepted — it's a remote document, and a name in it must never become a host Kunai sends requests to. There's a test with `evil.example.com` and `miruro.bz.attacker.test`.
   - `miruro.com` is excluded by name: the status page calls it a mirror, but it's a landing page with no pipe.
   - A mirror the status page calls **down is moved to the back, never dropped** — its vantage point isn't the user's.
3. **The mirror that last answered leads**, so a session pays once for finding one that works.

## Also

`curl` gets a `--connect-timeout`. The existing stall bound only starts counting once bytes flow, so a mirror whose TCP connect hangs held the full 25s `--max-time`. With more mirrors to get through, that's the difference between trying the next one and losing the attempt budget to one dead host.

The relay profile gains the two new mirrors and `status.miruro.com`. All of it is metadata; media stays direct.

## Verified

`test:live:miruro` still resolves and plays (`www.animegg.org`). `typecheck` 14/14 · `test` 23/23 · `lint` 0 warnings · `fmt` · `verify:doc-paths` — all `--force`.

## Seams

- **Declaration → reader:** `mirrors.ts` is consumed by `pipeCall`; `recordMiruroMirrorSuccess` is called on every decodable body.
- **Reverse states:** a demoted mirror is still tried (last), and a status page that fails leaves the shipped list in place — both tested.
- **Every platform:** no platform-specific behaviour; `--connect-timeout` is supported by both plain curl and curl-impersonate.
- **Docs:** manifest notes corrected (the "TLS-dead" claim was wrong) and the mirror-order contract recorded.


KitsuneKode:
@coderabbitai review

KitsuneKode:
@coderabbitai review

KitsuneKode:
@coderabbitai review

KitsuneKode:
@coderabbitai review

PR #377, opened 2026-09-12
fix(videasy): stop the resolve gate verifying a request mpv never makes

### LINKED ISSUE #361: videasy: resolve gate attests reachable for a URL that 403s to mpv and every other client
Videasy attests `streamReachabilityVerified: true` for a URL that answers 403 to every other client, including mpv. The resolve looks green and playback cannot start.

Reproduced repeatedly on 2026-09-11 against fixture `bloodhounds` (`tmdb:127529` S1E2), isolated profile, `balanced`.

## What happens

The provider's own resolve gate probes the selected stream and gets `reachable`. Every subsequent request to that identical URL — with the identical headers, from the same process, seconds later — gets 403.

Gate, instrumented at `probeSelectedVidkingPayloadStream`:

```
[GATE] {"server":"wings-cdn","streamUrlTail":"...UXByaU55cDRsdw/playlist.m3u8",
        "headerKeys":["referer","origin","user-agent"],
        "healthy":true,"probed":true,"probe":{"status":"reachable"},"candidateCount":1}
```

Shipped result, same run:

```
{"streamReachabilityVerified":true,"selectedStreamId":"stream_1fzahluwtqpj","streamCount":1,
 "streams":[{"host":"moon.peakstorm.top","tail":"UXByaU55cDRsdw/playlist.m3u8"}]}
```

Same run, moments later, same URL and headers:

```
runStreamHealthCheck(resolve-gate) -> {"healthy":false,"probe":{"status":"unreachable","reason":"HTTP 403","definitive":true}}
probeStreamReachability            -> {"status":"unreachable","reason":"HTTP 403","definitive":true}
mpv --no-config --vo=null --ao=null --frames=2 --ytdl=no --referrer=... --user-agent=... --http-header-fields=Origin:...
                                   -> exit 2, "Failed to open ...playlist.m3u8"
```

curl, every combination, both plain and impersonate:

```
curl            no headers        -> 403      curl_chrome150  no headers        -> 403
curl            referer only      -> 403      curl_chrome150  referer only      -> 403
curl            referer+UA        -> 403      curl_chrome150  referer+UA        -> 403
curl            referer+UA+origin -> 403      curl_chrome150  referer+UA+origin -> 403
```

## What it is not

Ruled out by experiment, so please don't re-file these:

- **Not headers.** 403 with and without every combination the provider attaches.
- **Not TLS fingerprinting.** `curl_chrome150` behaves identically to plain curl.
- **Not the `ytdl_hook` artifact.** A first attempt without `--ytdl=no` produced a misleading `[ytdl_hook] HTTP Error 403`; re-run with `--ytdl=no` (what `mpv.ts:551-559` does for remote HLS) still fails.
- **Not the gate consuming a one-shot URL.** With the gate short-circuited so it issues no request at all, the first fetch is still 403.
- **Not a per-resolve token.** `selectedStreamId` and the URL tail are byte-identical across separate runs with fresh isolated profiles, so the URL is deterministic for this title/route.

## What is still open

The gate's fetch gets 200 and the very next identical fetch gets 403, with no cookie jar involved (`createRelayFetchPort` delegates straight to `fetch` when `baseUrl` is empty, and nothing in `videasy/` touches cookies). The most consistent reading is that the CDN grants one fetch per window against a deterministic URL, and whichever client asks first spends it — which in production is always our own gate. That is not proven; distinguishing it from a very short TTL needs a clean window this session could not get after repeated probing.

## Why it matters

`streamReachabilityVerified: true` is trusted downstream — `shared/stream-health.ts:163,177,208` gate on `=== true`, and `playbackTrustMs` is 5m — so a false green is also cached and replayed. Videasy is the configured default (`packages/config/src/defaults.ts:26`), so this is the first thing a new user hits.

It also suppresses failover: only one candidate was produced, and because the gate passed, `probeSelectedVidkingPayloadStream`'s caller never reached its `break` and never tried another flavor.

## Suggested direction (needs a decision, not a drive-by)

1. **Stop the false green first.** The gate may keep failing fast on a definitive failure, but it should not report `verified: true` from a probe whose success does not survive to the player. Downgrading to `undefined` is honest and stops the 5m replay. Low risk, no failover change.
2. **Then make the check represent playback.** A URL that serves one request and refuses an immediate identical second cannot support a player, which makes many. Probing twice and requiring both would classify this correctly and let the loop `break` to the next flavor. Costs one extra request per candidate for the two providers using the shared gate, so it wants a measurement before it lands.

`knownGoodLabels` for this fixture are Neon/Cypher/Yoru; whether those play once failover is reachable is untested.

## Reproducing

Temporary probe used here is not committed. It resolves through `resolveProviderSmokeStream`, prints the shipped URL/headers, runs `mpv --frames=2` with `--ytdl=no` for HLS, then re-runs the gate predicate and curl variants against the same URL. `bun run --cwd apps/cli test:live:videasy` shows the same end state as a hard check: `Stream not playable (resolved=true, reachable=false)`.

KitsuneKode:
Closes #361.

## What was wrong

Videasy's resolve gate probed the selected stream and shipped `streamReachabilityVerified: true`, then mpv could not open the URL at all. The gate and the player were not asking the same question:

| | `Origin` sent |
| --- | --- |
| resolve gate (`probeSelectedVidkingPayloadStream`) | `https://www.vidking.net` — the default in `normalizeStreamCandidates` |
| the stream that ships | `https://www.cineby.at` — `clientProfile.origin` |

The CDN behind the **Yoru** server accepts the first and answers **403** to the second. So the gate verified a request that was never made, and the false green was then cached (`playbackTrustMs` 5m) and replayed.

The divergence dates to the rename commit that made `bc-frontend` the default profile: the profile's API origin was applied to the stream too, and the gate kept its own default.

## How it was pinned down

Asking **one URL twice in the same second**:

```
200  vidking-origin + cineby-referer  (what the gate sends)
403  cineby-origin  + cineby-referer  (what ships)
200  vidking-origin + vidking-referer
200  no origin      + cineby-referer
```

Same result on two separate titles. The CDN objects to exactly one thing: `Origin: https://www.cineby.at`.

That also **refutes the three hypotheses in the issue**, so they don't get re-investigated:

- **Not a one-shot URL.** With the gate stubbed so it never touched the CDN, mpv going first still got 403.
- **Not a short TTL.** Bun repeating the gate's own headers got 200 for 20+ seconds, and the gate's own second request succeeded at t+2.5s.
- **Not a client fingerprint.** Bun, curl, curl-impersonate and mpv all get 403 with the shipped header and 200 with the gate's.

The issue's "not headers" check was sound but only varied the *shipped* headers — never the gate's.

## The fix

`streamOrigin` becomes one field on `VideasyClientProfile` — the vidking.net player every site in this family embeds, which is what a browser's CDN request actually carries — and both the gate and the shipped result read it. The **API** request keeps the site's own origin (`cineby.at`), which is what it wants.

The gate's `streamOrigin` parameter is **required**, so a future call site cannot silently fall back to the default and drift apart again. That structural point is the real fix; the value is just the value.

## Verification

Live against the provider, not only unit tests:

- Fresh resolve → **mpv writes a frame** under `--vo=image` for **Dune** (movie, 4K) and **Bloodhounds S1E2** — the issue's own fixture, which never opened before.
- The other CDNs in this family (Vyse, `i-arch-*.gufin435siv.com`) accept `vidking.net`, `cineby.at` and no `Origin` alike, so they are unaffected.
- Full `--force` gates green.

## Noted, not fixed here

Two separate things seen while testing, neither caused by this change: a resolved peakstorm URL stops working after tens of minutes (a stale-URL lifetime, the class already noted for cached streams), and mpv is sometimes slow to open on the `sun.peakstorm.top`/`Breach` server. Neither blocks a fresh play.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Videasy playback failures on affected Yoru servers by aligning provider verification with the origin used for embedded streams.
  - Improved stream handling for both supported Videasy app configurations while preserving behavior for other servers.

- **Tests**
  - Added coverage confirming streams use the player’s origin rather than the API origin.

### DIFF
--- a/.changeset/videasy-stream-origin.md
+++ b/.changeset/videasy-stream-origin.md
@@ -1,0 +1,16 @@
+---
+"@kitsunekode/kunai": patch
+---
+
+Fix Videasy shipping streams that could not be opened, on the server it picks most often.
+
+Videasy checked each candidate stream before offering it, and reported the check
+had passed — then playback failed immediately. The check and the player were not
+asking the same thing: the check sent one `Origin` header and the stream Kunai
+handed to mpv carried another, and the CDN behind Videasy's Yoru server accepts
+the first and refuses the second. So the provider verified a request that was
+never made, and the failure only appeared once the player had already started.
+
+Both now send the origin of the player these sites actually embed, taken from
+one place so they cannot drift apart again. Where this was failing it now plays;
+the other Videasy servers accept either header and are unaffected.
--- a/packages/providers/src/videasy/direct.ts
+++ b/packages/providers/src/videasy/direct.ts
@@ -138,9 +138,20 @@ export function displayHostForEndpoint(endpoint: string): string {
 
 export type VideasyClientProfile = {
   readonly appId: string;
+  /** Origin sent to the sources API — the site fronting it. */
   readonly origin: string;
   readonly defaultReferer: string;
   readonly streamReferer: string;
+  /**
+   * Origin sent with the stream itself, which is not the API's. Every site in
+   * this family plays media inside the vidking.net player, so that is the origin
+   * a CDN sees from a browser. The peakstorm CDN behind the Yoru server refuses
+   * `Origin: https://www.cineby.at` with a 403 while accepting vidking.net or no
+   * Origin at all (2026-09-12); the other CDNs accept either. The resolve gate
+   * and the shipped stream both read this one field — they diverged once, and
+   * the gate spent months verifying a request mpv never made (issue #361).
+   */
+  readonly streamOrigin: string;
 };
 
 const USER_AGENT =
@@ -1280,6 +1291,7 @@ async function probeSelectedVidkingPayloadStream({
   sourceId,
   server,
   streamReferer,
+  streamOrigin,
   sourceQualityFilter,
   engineOptions,
   events,
@@ -1292,6 +1304,8 @@ async function probeSelectedVidkingPayloadStream({
   readonly sourceId: string;
   readonly server: VidkingServerEndpoint;
   readonly streamReferer?: string;
+  /** Must be the shipped stream's origin: the gate verifies the request mpv will make. */
+  readonly streamOrigin: string;
   readonly sourceQualityFilter?: string;
   readonly engineOptions?: VidKingEngineOptions;
   readonly events: ProviderTraceEvent[];
@@ -1306,6 +1320,7 @@ async function probeSelectedVidkingPayloadStream({
     sourceId,
     server,
     streamReferer,
+    streamOrigin,
     sourceQualityFilter,
     flavorLabel: presentation.themeLabel,
     serverName: presentation.themeLabel,
@@ -1582,6 +1597,7 @@ async function tryVidkingServer(opts: {
   const appId = clientProfile.appId;
   const requestReferer = customReferer ?? clientProfile.defaultReferer;
   const streamReferer = clientProfile.streamReferer;
+  const streamOrigin = clientProfile.streamOrigin;
   const requestServers = resolveVideasyRequestServers(server, clientProfile);
 
   emitTraceEvent(events, context, {
@@ -1738,6 +1754,7 @@ async function tryVidkingServer(opts: {
             sourceId,
             server,
             streamReferer,
+            streamOrigin,
             sourceQualityFilter: engineOptions.filterQuality,
             engineOptions,
             events,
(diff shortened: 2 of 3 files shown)


KitsuneKode:
Addressed in 595c1c14, though not with the requested test — the resolve path decrypts the sources payload through WASM, so a unit test cannot drive `tryVidkingServer` to the probe without an encrypt side.

The concern is right, so it is fixed at the type level instead, which catches strictly more: `normalizeStreamCandidates` is the only place stream headers are built and both paths come through it, so its `streamReferer`/`streamOrigin` are now **required** rather than optional-with-VIDKING-defaults. Optional defaults are precisely how the two drifted — the gate took the defaults, the result took the profile. Omitting either at either call site is now a compile error, and both sites read the same profile pair in one scope. (`createVidkingResultFromPayload` stays callable without them and applies the fallbacks at its own call.)

End-to-end behaviour is verified live instead of in a unit test: fresh resolves write an mpv frame for Dune and for the issue's `bloodhounds` S1E2 fixture, which never opened before.

PR #382, opened 2026-09-12
fix(providers): read a JSON body at the boundary, not through a cast

KitsuneKode:
Found while verifying every production provider end to end.

## What happens

VidLink answers **`200 OK` with the body `null`** for every title right now. Kunai did:

```ts
const data = (await response.json()) as { stream?: VidlinkStream };
const stream = data.stream;   // TypeError: null is not an object
```

So a provider saying "I have nothing" surfaced as an error from inside resolve:

```
vidlink exhausted 1885ms  null is not an object (evaluating '(await (await fetchVidlinkApi(ctx, `${VIDLINK_API_BASE}…
```

The lane blamed itself instead of stepping aside, and the real message ("no sources") never appeared.

## Scope — four providers, same shape

Five call sites dereferenced a body that can be JSON `null`:

| Provider | Call |
| --- | --- |
| VidLink | sources (`data.stream`) — **failing live today** |
| Videasy | the Wings seed (`body.seed`) and the TMDB proxy (`data.first_air_date`) |
| AllManga | the AniList title lookup (`payload.data`) and the key bootstrap (`body.partB`) |

All now go through `readJsonObjectBody`, which returns the object (or an array) and `null` for `null`/scalars, at the one place that holds the raw body — so callers branch on the domain value instead of trusting a cast. A body that isn't JSON at all still throws, because that *is* a broken response.

## Verification

- Live: vidlink now ends with `VidLink returned no playable streams` instead of a TypeError.
- Regression test in `vidlink-dash.test.ts` replays the exact live reply (`200` + `null`) and asserts the result is `exhausted`, with no `is not an object` in any failure message.
- Full `--force` gates plus `build` green.

## What this does not do

VidLink still has no sources — that is upstream (checked with and without the `x-playback-environment` header; the encryptor at enc-dec.app is fine). This only makes the failure honest and the fallback immediate.

### DIFF
--- a/.changeset/provider-null-json-body.md
+++ b/.changeset/provider-null-json-body.md
@@ -1,0 +1,14 @@
+---
+"@kitsunekode/kunai": patch
+---
+
+Stop a provider that answers "nothing" from failing as an internal error.
+
+Some provider APIs reply `200 OK` with a body of `null` when they have no
+source for a title — VidLink does it for every title right now. Kunai read a
+field off that reply and threw, so instead of moving on to the next provider it
+reported an error from inside itself.
+
+Four providers read a response this way — VidLink, Videasy, AllManga's title
+lookup and its key bootstrap — and all four now treat an empty reply as "no
+source here" and hand over to the next one.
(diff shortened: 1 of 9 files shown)


KitsuneKode:
@coderabbitai review

KitsuneKode:
@coderabbitai review

KitsuneKode:
@coderabbitai review

PR #383, opened 2026-09-12
fix(miruro): treat a rate-limited backend as unusable, not as a candidate

KitsuneKode:
Stacked on #368. Review the last commit only.

## Found by the release gate, not by hand

`KUNAI_LIVE_RELEASE_SIGNOFF=1 bun run test:live:release-signoff` failed on the anime lane:

```
{ "lane": "anime", "configuredProvider": "miruro", "successfulProvider": "miruro",
  "resolved": true, "streamReachable": false, "failureClass": "provider-drift" }
```

Miruro returned exactly one candidate — its `pewe` server, whose backend is `hls.anidb.app`, the one that has been in maintenance. It answered **429** to:

| client | status |
| --- | --- |
| mpv | 429 |
| curl with the stream's headers | 429 |
| curl with no headers | 429 |

So the lane resolved a stream no player could open. The user would have watched the player open and fail, with recovery only starting after that.

## Why 429 was excluded, and why that was wrong

`isMiruroBackendDownStatus` deliberately ignored 401/403/429, because **owocdn behind `kiwi` answers Bun's fetch 403 while mpv plays the same URL** — a status only a non-player sees. That reasoning is sound for 403 and does not transfer to 429: a rate-limited master is refused to *everyone*, as all three clients above agree.

Trade: rejecting costs one more server attempt (the probe is a single ranged GET with a 2s timeout); accepting cost a failed play on the **default anime provider**. 401/403, plain 500 (AnimeGG's vidcache hand-off) and cross-host redirects are still never rejections.

## Verification

- With `pewe` skipped, the cycle lands on `moo` and **mpv writes a 1080p frame** (Frieren ep 1).
- The signoff now reports `ok: true` with all three lanes `streamReachable: true` (movie 2.1s, series 2.2s, anime 13.1s — anime pays one 2s probe to skip the dead server).
- Full `--force` gates green; 122 provider tests pass, including a new one pinning 429 as a rejection and keeping 401/403/416 as non-evidence.

Also verified in the same pass, unchanged by this PR: YouTube (`ok: true` for resolve, a 2160p ladder and `type:short` search; "Me at the zoo" decodes through mpv + yt-dlp), AnimeGG direct (frame written), and KickAssAnime (frame written).


KitsuneKode:
@coderabbitai review

PR #386, opened 2026-09-12
fix(miruro): stop paying for backends that cannot answer

### LINKED ISSUE #384: miruro: re-probes a known-down backend on every resolve (anime lane 13s vs 2s)
Miruro is the only production provider that does not use `EndpointHealthPort`, so it re-probes a backend it already knows is refusing, on every resolve.

## Measured

With `pewe`'s backend (hls.anidb.app) rate-limited, on the branch that rejects a 429 candidate (#383):

| lane | resolve |
| --- | --- |
| movie (videasy) | 2.1 s |
| series (videasy) | 2.2 s |
| **anime (miruro)** | **13.1 s** |

The anime lane pays a 2 s `probeMiruroBackendDown` timeout on `pewe`, discards it, then resolves from `moo` — and pays it again on the next episode, for as long as the backend stays down. `grep -c endpointHealth packages/providers/src/miruro/direct.ts` → `0`.

## Fix direction

Follow `createVideasyEndpointHealth` (`videasy/direct.ts:196`), which already wraps the port with an in-process fallback:

1. In the cycle's `resolveCandidate` (`miruro/direct.ts`, the `downStatus !== null` branch), `recordFailure(metadata.serverId, { class: "server-error", titleId })` before throwing, and `recordSuccess(metadata.serverId)` on the returning path.
2. Consult `shouldTry(serverId)` when candidates are planned, so a quarantined backend is skipped without a probe.

Two constraints, both learned the hard way here:

- **Only a definitive backend status may be recorded** — the 404/410/429/502/503/504 from the same host. A timeout, a cancel, or an aborted hedge loser must never poison health; that previously produced a 1-hour persisted quarantine for servers that were fine.
- **Never quarantine the whole list into failure.** If `shouldTry` rejects every server, try them anyway — a self-inflicted total outage is worse than one slow probe.

## Not urgent

The lane works and plays (verified with a written 1080p frame); this is startup latency on the default anime provider while an upstream backend is down, not a failure.

KitsuneKode:
Stacked on #383. Closes #384 — and corrects it.

## #384 guessed wrong

The issue blamed a 2 s probe timeout on the dead `pewe` backend. A 429 answers in ~300 ms. Timing every request in one resolve against the live pipe:

| cost | cause |
| --- | --- |
| **~2000 ms** | the **`moo`** probe timing out — on every resolve |
| ~1300 ms | three Bun fetches Cloudflare refuses before curl answers |
| ~650 ms | `pewe`'s ladder GET + probe |

## Moo: a probe that could never answer

Every `moo` stream starts at `animegg.org/play/…`, and that endpoint **never answers Bun's fetch** — 6 s and still nothing, with or without following redirects — while curl and mpv get a 302 immediately. The probe returns `null` on a timeout, so it spent its full 2 s to learn nothing, on the backend that actually plays now. It no longer asks `animegg.org`; AnimeGG's own dossier already says a probe cannot judge its streams.

## Pewe: asked again every episode

Miruro was the only production provider not using `EndpointHealthPort`. The shared cycle engine already skips quarantined candidates — Miruro just never passed the port. Now it does, and a definitive backend status is recorded as `server-error` against the server id (`pewe`, `moo`; sub and dub share a backend).

The engine deliberately records nothing for `candidate-empty`, because an episode a server lacks is not evidence against it — so Miruro records the definitive status itself, and **never** for a timeout or an aborted probe. That keeps clear of the earlier incident where an over-eager record quarantined healthy servers for an hour.

## Measured

Real `ProviderEndpointHealthService`, in-memory store, live pipe, consecutive episodes:

| | resolve |
| --- | --- |
| before | ~4.3 s |
| with the Moo fix | ~2.3 s |
| once `pewe` is quarantined | **~1.1 s** |

The first play still varies (9–30 s seen) depending on which backends the catalog offers in that fetch — upstream, not this change.

## Tests

`miruro-endpoint-health.test.ts` drives the **real cycle against a stubbed pipe** (XOR-encoded responses), so it pins behaviour a user sees:

- a rate-limited backend is recorded and the next one plays
- a quarantined backend is **not asked at all**
- a probe that couldn't reach the backend records nothing
- an episode a server lacks records nothing
- a server that plays clears its record

Three of those fail against the unfixed code; the two "no false evidence" guards pass both ways, as they should. Probe tests that used an AnimeGG URL were repointed so the 500 and cross-host rules they name are still actually exercised.

## Not done, on purpose

The ~1.3 s of Cloudflare round trips before curl. Skipping straight to curl would need a process-level memo in the pipe transport that #358 hardened; ~0.5 s isn't worth that risk here.

## How to see it and how it ends

A set-aside backend shows in the source list as "temporarily unavailable" and is retried when its hour is up — the existing behaviour for Videasy. There is no manual reset for endpoint health in the CLI today; that's pre-existing and unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Miruro playback startup by skipping backends that cannot be evaluated promptly.
  * Temporarily sets repeatedly failing backends aside for one hour, preventing repeated delays between episodes.
  * Clearly marks temporarily unavailable backends in the source list and retries them after the waiting period.
  * Avoids probing a consistently unresponsive endpoint, reducing stream resolution delays.
  * Added a playback fallback check to better confirm stream availability when standard reachability checks are inconclusive.

### DIFF
--- a/.changeset/miruro-faster-start.md
+++ b/.changeset/miruro-faster-start.md
@@ -1,0 +1,16 @@
+---
+"@kitsunekode/kunai": patch
+---
+
+Start anime from Miruro in about a second instead of about four.
+
+Two things were costing every episode. Kunai checked each Miruro backend before
+playing it, and for the one that plays most reliably right now that check could
+never get an answer — so it waited out its full timeout every time, for nothing.
+And a backend that was down got checked again on every single episode.
+
+The check now skips the backend it cannot judge, and a backend that keeps
+failing is set aside for an hour instead of being asked again, the same way
+Kunai already does for its movie and series sources. While it is set aside the
+source list shows it as temporarily unavailable, and it is tried again once the
+hour is up.
--- a/.docs/providers.md
+++ b/.docs/providers.md
@@ -533,6 +533,20 @@ demands — the stream does not play for anyone — so Miruro now matches the
 resolve gate's policy that every 4xx is definitive, and 401/403 remain the
 narrow exception because only Bun's fetch sees them.
 
+A rejected backend is recorded against its server id (`pewe`, `moo` — sub and
+dub share one) as a `server-error` through `context.endpointHealth`, and the
+cycle is handed the same port, so a backend quarantined by the usual rules
+(three consecutive failures on one title, or two titles) is skipped without a
+request. The engine records nothing for `candidate-empty`, which is correct — an
+episode a server lacks is not evidence against the server — so the definitive
+status is recorded by Miruro itself, and never for a timeout or an aborted probe.
+
+The probe does not ask `animegg.org` at all. Its `/play` endpoint, where every
+`moo` stream starts, never answers Bun's fetch while curl and mpv get a 302 at
+once, so each probe spent the full timeout to learn nothing. Measured on
+2026-09-12 against the live pipe, a Miruro resolve went from ~4.3 s to ~2.3 s
+from that alone, and ~1.1 s once `pewe` was quarantined.
+
 Miruro's own single point of failure is `miruro.bz`/`.ru` — every one of its
 backends is reached through it, so the two providers behind it are the ones that
 share none of that. `kickassanime` is second and `animegg` third: both have their
--- a/.docs/release-reliability-gate.md
+++ b/.docs/release-reliability-gate.md
@@ -261,6 +261,16 @@ Expected result for each provider:
 - `failureCodes` is empty or contains only non-blocking fallback evidence when a fallback stream was selected
 - output includes `isolatedProfile: true`
 
+The route signoff needs `streamReachable: true` — proof, not neutrality. Its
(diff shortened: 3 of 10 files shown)


KitsuneKode:
Added a30fe76b after running the release signoff on every fix combined: with `pewe` skipped the anime lane lands on `moo`, whose `animegg.org/play` never answers Bun's fetch, so the signoff's Bun probe timed out → `streamReachable: null` → `provider-drift` → **red gate on a stream mpv plays**. The signoff keeps its `=== true` bar; a timed-out probe is now settled by mpv decoding one frame. Result on the integration branch: `ok: true`, all three lanes reachable, with `[signoff] anime: probe inconclusive; mpv decode says true` on stderr.

KitsuneKode:
@coderabbitai review

PR #392, opened 2026-09-16
fix(providers): make provider-health reset clear endpoint quarantines

KitsuneKode:
Stacked on #390 - review that first; this branch's own commit is the reset fix.

## What was wrong

Quarantined endpoints (1h `server-error`, 24h `route-dead` in `provider_endpoint_health`) had **no manual escape**: `applyProviderHealthResetScope` cleared global `providerHealth` and per-title `titleProviderHealth` but never touched `container.endpointHealth`. After a reset the UI said retry, the cycle kept emitting `source:skipped … quarantined`, and the user had to wait out the hour. The reset confirmation was a lie for endpoint-skipped candidates.

## Fix

- `ProviderEndpointHealthService`: new `delete` / `deleteByProvider` / `clearTitle(titleId, providerId?)` / `clearAll` - every one also drops in-memory transient cooldowns and curated dead seeds so no ghost state survives a reset.
- `ProviderEndpointHealthRepository`: new `deleteByProvider`.
- `provider-health-reset.ts`: every scope clears the endpoint rows it owns (provider, lane, all). Per-show scopes only lift rows the title contributed evidence to (`distinctTitleIds`); rows with no title evidence stay quarantined. Return gains `clearedEndpoints` (single caller ignores it) and the confirmation names the lifted count.

## Verified

- Extended `provider-endpoint-health-service.test.ts` (deleteByProvider isolation, clearTitle title-scoping, clearAll incl. transient cooldowns) and `provider-health-reset.test.ts` (scopes clear endpoint rows; existing assertions unchanged).
- `apps/cli` + `packages/storage` typecheck clean; 219 playback unit tests pass; oxlint clean.


KitsuneKode:
## CI note: macOS failure is the known installer wall-clock flake (#385), not this PR

The single failing job (`macOS CLI parity`) fails only these two tests:

- `install.sh lifecycle contract > bounds activation contention by real elapsed time when poll exceeds timeout` (674ms vs 300ms budget)
- `install.sh lifecycle contract > a failed reclaim with zero poll observes the activation deadline instead of hot-looping` (556ms vs 500ms budget)

Why it is not this PR:

- This stack touches `packages/core`, `packages/providers`, `services/playback`, `packages/storage`, docs and changesets. **Zero installer files** (`git diff --name-only main...fix/provider-endpoint-reset` above - no `install.sh`, no installer tests).
- The same two tests failed on #382, another providers-only PR - documented in #385, which fixes them by expressing the claim as a ratio instead of an absolute wall-clock budget.
- Both pass locally (113-155ms), well inside budget on an unloaded machine. The macOS runner exceeded the budget with the lock logic behaving correctly - exactly the flake profile.

I re-ran the failed job to confirm it goes green on retry. Merging #385 removes the flake for every PR.

